PR #3, opened 2020-04-27
Addition of the "endomorphism trick" in the code generation. Code is generated using Bavard now.

ThomasPiellard:
internal/: I moved around the folders and cleaned the code generation, it uses bavard now. Currently there are no command line tools to generate curves but we can add one if needed.
The code generation for the pairing is cleaned up as well, all the curve specific parts are correctly handled now (like the mulAssign for the type lineEval).

<curves>/: I added the endomorphism trick (in g1.go), it's faster than a normal scalar multiplication but require the scalar to be correctly split. It returns a chan like the multiExp.
The signature is:
```
// ScalarMulEndo performs scalar multiplication
// using the endo phi(p=(x,y))=(ux,y) where u is a 3rd root of 1,
// phi(P) = lambda*P
// u = 4002409555221667392624310435006688643935503118305586438271171395842971157480381377015405980053539358417135540939436
// lambda = 228988810152649578064853576960394133503 (128 bits)
// s1, s2 are scalars such that s1*lambda+s2 = s
// s1 on 128 bits
// s2 on 128 bits
func (p *G1Jac) ScalarMulEndo(curve *Curve, a *G1Affine, s1, s2 fr.Element) chan G1Jac
```


gbotrel:
1. Please fix CircleCI build :-) 
2. Can you find a way to have a single `ScalarMul` op. that has the same signature? And returns a points instead of a chan? Seems a bit confusing to have 2 similar APIs, one synchronous, the other not. 
3. Is splitting `s` into `s1` and `s2` costly? Is it meant to be done "on the fly" ? (if so, then I'ld prefer to have a single `ScalarMul` op, not `ScalarMulEndo`

ThomasPiellard:
Ok I'll change the signature to return a point. 

The time to find s1 and s2 is the time of a euclidean division, either we use big.Int or maybe you can add this in goff (it's why I wanted it in the first place), I expect it not to take too long... If the overall time is lower we should have only this scalar mul implemented. Do you think that I should implement it for G2 as well?

Btw I only implemented this for G1 because it's for the verifier that it might be the more useful, but it would work for G2 as well if needed. Without the time for splitting s into s1 and s2 the method is faster (by a non negligible factor).

bls377/g1.go
```diff
@@ -469,6 +486,113 @@ func (p *G1Jac) Double() *G1Jac {
...
+// u = 80949648264912719408558363140637477264845294720710499478137287262712535938301461879813459410945
+// lambda = 91893752504881257701523279626832445440 (129 bits)
+// s1, s2 are scalars such that s1*lambda+s2 = s
+// s1 on 129 bits
+// s2 on 127 bits
+func (p *G1Jac) ScalarMulEndo(curve *Curve, a *G1Affine, s fr.Element) G1Jac {
+
+	// operation using big int
+	var lambda, _s, _s1, _s2 big.Int
+	lambda.SetString("91893752504881257701523279626832445440", 10)
```
ggutoski, line 527: Why are we doing string-to-integer conversion of a constant in every single call to `ScalarMulEndo`?  (Same for line 536.)  We should hard-code this constant instead.  Even if we are stuck using `big.Int` there _must_ be a faster way to accomplish this.

Is `big.Int` really the best choice here?  Seems to me that `fr` is preferable.  Regarding your [earlier comment](https://github.com/ConsenSys/gurvy/pull/3#issuecomment-621368191): seems worth it to add Euclidean division to `goff` for use here.  Perhaps we should open a new `goff` issue for this task.

bls377/g1.go
```diff
@@ -496,6 +620,86 @@ func (p *G1Jac) ScalarMulByGen(curve *Curve, scalar fr.Element) *G1Jac {
...
+	// Store only one set T_i at a time---don't store them all!
+	var T [TSize]G1Jac // a set T_i from Bootle, the set of g^j for j in [1:2^b] for some choice of g
+	computeT := func(T []G1Jac, t0 *G1Jac) {
+		T[0].Set(t0)
+		for j := 1; j < (1<<b)-1; j = j + 2 {
+			T[j].Set(&T[j/2]).Double()
+			T[j+1].Set(&T[(j+1)/2]).Add(curve, &T[j/2])
+		}
+	}
+	return p.pippenger(curve, points, scalars, s, b, T[:], computeT)
```
ggutoski, line 663: Not part of this PR, but oh man we really need to clean up this part of multiexp.  That `computeT` parameter is ugly. 😝

bls381/notes
```diff
@@ -0,0 +1,2 @@
+- bls381: we can speed up the scalar mul < 50 , via double & add, using the endo (x, y)->(u*x, y), where u is cube root of 1
```
ggutoski, line 1: Uh, maybe put this note into a github issue and delete this file? 😉

go.sum
```diff
@@ -10,8 +10,8 @@ github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghf
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/consensys/bavard v0.1.1 h1:sFtk3U10YO4Mt8NNCEZQhMzoQ9UMy2H5GKOx/DwNg80=
 github.com/consensys/bavard v0.1.1/go.mod h1:ffZkLPNQSN3E6u+zpArQSleJ/lsraMwKPCHQymPQJtM=
-github.com/consensys/goff v0.2.0 h1:a9nB5CGLbNXzUTflPzuqj1i96jv6vOAVbhlsEgx1nZw=
-github.com/consensys/goff v0.2.0/go.mod h1:RCji/8mD0qP8UZJCtEG4NEWD6JPwN7S4KBGvRXkVMgw=
+github.com/consensys/goff v0.2.1 h1:sBdsFEetlPpXnZJR0WtZaU6X+dDCejbHqG8M21WJtk8=
+github.com/consensys/goff v0.2.1/go.mod h1:CsKD9nM1/fD0gqJs0vRCyQ/wocVjex+wa3mVEjC6h+s=
```
ggutoski, line 14: as we discussed, please update to 0.2.2

bls377/g1.go
ThomasPiellard: Yep, we should do that in a separate branch I think...

bls377/g1.go
ThomasPiellard: It's not so bad performance wise. The algo is still faster than the normal scalar double and add. I'll for Gautam to implement euclidean division in goff, but I am not sure it's needed (at least it's not urgent for this algo).

Issue #13, opened 2020-10-05
Replace k/2-th Frobenius powers by conjugations


yelhousni:
`k/2`-th Frobenius powers can be replaced by conjugations in the final exponentiation of the pairing implementation for the curves. 

## Motivation
In Tate-based pairings, the target group `GT` is the `r`-th root of unity in `Fq^k` where `q` is the finite field prime, `r` the curve subgroup order and `k` the embedding degree. The pairing requires also a final exponentiation to `(q^k-1)/r` to reduce the output to a unique value. Using k-th cyclotomic polynomial, the exponent can be simplified into an easy part `(q^k-1)/phi_k(q)` and a hard part `phi_k(q)/r`. The easy part can be computed cheaply using Frobenius powers. However, if there is a power to `k/2` it should be replaced by a simple conjugation. An inverse is a power to `q^(k/2)` but since there is a `q^(k/2)-1` component in that exponent, the element `f` becomes unitary (i.e. norm 1) and the inversion becomes a simple conjugation.

PR #14, opened 2020-10-05
replace k/2-th Frobenius powers by conjugations

yelhousni:
Issue: #13 
Replace 6-th Frobenius power by unitary inverse in BN254, BLS12-377 and BLS12-381

**TODO**
Same goes for BW6-761 especially that it uses many `FrobeniusCube` in the hard part, but it seems that BW6 `GT`is implemented as `Fp3over2` rather than `Fp2over3`.

### DIFF
--- a/bls377/pairing.go
+++ b/bls377/pairing.go
@@ -49,8 +49,7 @@ func (z *GT) FinalExponentiation(x *GT) *GT {
 	var t [6]GT
 
 	// easy part
-	t[0].FrobeniusCube(&result).
-		FrobeniusCube(&t[0])
+	t[0].InverseUnitary(&result)
 	result.Inverse(&result)
 	t[0].Mul(&t[0], &result)
 	result.FrobeniusSquare(&t[0]).
--- a/bls377/pairing.go
+++ b/bls377/pairing.go
@@ -49,7 +49,7 @@ func (z *GT) FinalExponentiation(x *GT) *GT {
 	var t [6]GT
 
 	// easy part
-	t[0].InverseUnitary(&result)
+	t[0].Conjugate(&result)
 	result.Inverse(&result)
 	t[0].Mul(&t[0], &result)
 	result.FrobeniusSquare(&t[0]).
--- a/bls381/pairing.go
+++ b/bls381/pairing.go
@@ -52,8 +52,7 @@ func (z *GT) FinalExponentiation(x *GT) *GT {
 	var t [6]GT
 
 	// easy part
-	t[0].FrobeniusCube(&result).
-		FrobeniusCube(&t[0])
+	t[0].InverseUnitary(&result)
 	result.Inverse(&result)
 	t[0].Mul(&t[0], &result)
 	result.FrobeniusSquare(&t[0]).
--- a/bls381/pairing.go
+++ b/bls381/pairing.go
@@ -52,7 +52,7 @@ func (z *GT) FinalExponentiation(x *GT) *GT {
 	var t [6]GT
 
 	// easy part
-	t[0].InverseUnitary(&result)
+	t[0].Conjugate(&result)
 	result.Inverse(&result)
 	t[0].Mul(&t[0], &result)
 	result.FrobeniusSquare(&t[0]).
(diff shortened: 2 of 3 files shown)


mratsim:
If you're interested I've written the logic as comment in my own library: https://github.com/mratsim/constantine/blob/49164b66/constantine/pairing/cyclotomic_fp12.nim#L64-L76

```
Proofs:

Fp12 can be defined as a quadratic extension over Fp⁶
with g = g₀ + x g₁ with x a quadratic non-residue

with q = p⁶, q² = p¹²
The frobenius map f^q ≡ (f₀ + x f₁)^q (mod q²)
                      ≡ f₀^q + x^q f₁^q (mod q²)
                      ≡ f₀ + x^q f₁ (mod q²)
                      ≡ f₀ - x f₁ (mod q²)
hence
f^p⁶ ≡ conj(f) (mod p¹²)
Q.E.D. of (0)
```

mratsim:
Regarding the towering of BW6-761, the fact that Fp6 is implemented over Fp2 (cubic extension) shouldn't be a blocker. I have yet to implement BW6-761 (WIP) but you can compute the conjugate in a cubic extension with https://github.com/mratsim/constantine/blob/49164b6/constantine/tower_field_extensions/cubic_extensions.nim#L210-L222

```Nim
func conj*(a: var CubicExt) {.inline.} =
  ## Computes the conjugate in-place
  a.c0.conj()
  a.c1.conjneg()
  a.c2.conj()

func conj*(r: var CubicExt, a: CubicExt) {.inline.} =
  ## Computes the conjugate out-of-place
  r.c0.conj(a.c0)
  r.c1.conjneg(a.c1)
  r.c2.conj(a.c2)
```

with conjneg the negated conjugate. This works for me on Fp2 -> Fp4 -> Fp12 towering for classic curves (BN254-Snarks, BLS12-377 and BLS12-381)

bls377/pairing.go
```diff
@@ -49,8 +49,7 @@ func (z *GT) FinalExponentiation(x *GT) *GT {
 	var t [6]GT
 
 	// easy part
-	t[0].FrobeniusCube(&result).
-		FrobeniusCube(&t[0])
+	t[0].InverseUnitary(&result)
```
ThomasPiellard, line 52: Can you replace `t[0].InverseUnitary(&result)` by `t[0].Conjugate(&result)`? Because at this stage the element is not unitary, we just want to apply the Frob^6, which indeeds acts as a conjugation on Fp12 (I would like to keep `InverseUnitary` for elements which are really unitary, even if at the end it's the same operation). Same for the other curves :)

yelhousni: Makes sense :)

ThomasPiellard:
@mratsim what polynomials do you use for Fp2->Fp4-Fp12 on e.g. bn254? Because to know how let's say Frob^6 acts on Fp12, I would write y in Fp12 as y=a+bx+cx**2, where a, b, c are in Fp4. Here Frob6 acts as a conjugation on a, b, c (over Fp2), but to know how it acts on x I would write x^(p**6)=x*(x^(p**6-1)/4)**4=x*(x**4)**(p**6-1/4), where x**4 is in Fp3 (if the minimal polynomial of x is of the form x**3-d, which I assume it's the case). In your example it seems that you know directly a simple expression of Frob6(x)

ThomasPiellard (approved):

mratsim:
I use the polynomial described in Pairing implementation revisited: 
- https://eprint.iacr.org/2019/077.pdf
```
> An Fp12 value can be represented as Fp2[ω]/(ω^6−ξ), that is as an array of six Fp2 elements A+Cω+Eω^2+Bω^3+Dω^4+Fω^5 which can be written as (A+Bω^3) + (C+Dω^3)ω+ (E+Fω^3)ω^2 if we consider an Fp12 as a triplet of Fp4 elements
```
- The hard part of final exponentiation for BN curves follows Fuentes-Castaneda: https://github.com/mratsim/constantine/blob/244f583/constantine/pairing/pairing_bn.nim#L165-L197

ThomasPiellard:
Ok yes, indeed Frob^6(w) = -w, sorry I didn't see it (I didn't remember the degree 6 polynomial was sparse). So for BW6 it's the same , with Fp->Fp6=Fp(u) where u^6 is in Fp, Frob^3(u)=-u, so it's just a matter of flipping signs. I'll add this to bw6 for the final expo

PR #25, opened 2021-01-07
fp12: improved cyclotomicSquare

gbotrel:
see fq12.go tempalte --> removed one `MulByNonResidueInv` (cancelling with `MulByNonResidue`) and used E2 squaring (significantly faster than e2 mul) in fp4square . 

on my machine: 
```
benchmark              old ns/op     new ns/op     delta
BenchmarkPairing-8     903356        827455        -8.40%
BenchmarkPairing-8     907516        823723        -9.23%
```


yelhousni (approved):
LGTM

PR #27, opened 2021-01-08
Pairing: remove preCompute and go routines in MillerLoop

gbotrel:
@yelhousni following your recent awesome PR (#26) did some experiments and benchmarks on various architectures (x64), that confirms that for a small number of pair of points in the `Pairing` function, there is a severe penalty (specifically on `amd` platforms). 

In this PR, I just removed the go routines, the multiple `[68]lineEvaluation` (replace by a single `lineEvaluation` object) and merged the `preCompute` loop within the line evaluation / result compute loop. Did on `bn256` and `bls381`.

** If we want to deal with large number of pairs in `Pairing`, then we should probably revisit the `preCompute` strategy ** 

Here is where we're at on our reference benchmark (running on AWS z1d.3xlarge instance, with hyperthreading disabled) :

 ||mcl(ns/op)|gurvy(ns/op)|
| -------- | -------- | -------- |
|Pairing(bn256)	|478244	|489258|
|Pairing(bls381)	|676513	|707984|

For the before/after removal of go routines and precompute:

on AMD Ryzen 3700X, bn256 and bls381:
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkPairing-16        717847        429341        -40.19%
BenchmarkMillerLoop-16     330055        209102        -36.65%
```
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkPairing-16        1156689       644036        -44.32%
BenchmarkMillerLoop-16     517074        308647        -40.31%
```

on Intel MBP, less conclusive but not too bad:
```
benchmark                 old ns/op     new ns/op     delta
BenchmarkPairing-8        561681        546137        -2.77%
BenchmarkMillerLoop-8     253963        263012        +3.56%
```

### DIFF
--- a/bn256/pairing.go
+++ b/bn256/pairing.go
@@ -16,7 +16,6 @@ package bn256
 
 import (
 	"errors"
-	"sync"
 
 	"github.com/consensys/gurvy/bn256/internal/fptower"
 )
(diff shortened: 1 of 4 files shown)


bn256/pairing.go
```diff
@@ -122,137 +121,83 @@ func FinalExponentiation(z *GT, _z ...*GT) GT {
...
-		for i := 0; i < len(evaluations); i++ {
-			lineEvalPool.Put(evaluations[i])
-		}
-	}()
-
 	return result, nil
 }
 
-func lineEval(z *GT, l *lineEvaluation, P *G1Affine) *GT {
+func lineEval(result *GT, l *lineEvaluation, P *G1Affine) *GT {
```
yelhousni, line 194: We can maybe inline `lineEval()` as in `bls381/`, which makes the Miller Loop on a par with `develop` branch (w/ goroutines).

yelhousni (approved):
LGTM

bn256/pairing.go
yelhousni: resolved in 5ab1d22f07799ac023679d42093b265e5c1e58cb.

yelhousni:
In the last commit, I applied the same changes to `bls377`.

PR #42, opened 2021-04-21
Feat/fiat shamir

ThomasPiellard:
This PR adds a transcript object to apply Fiat Shamir transform.

## API breaking change

Addition of method `Bytes() []byte` on Digest struct:
```
type Digest interface {
	io.WriterTo
	io.ReaderFrom
	Bytes() []byte
}
```

## Transcript

To apply Fiat Shamir one needs to create a transcript while providing the hash function used for challenges derivation as well as a list of names for the challenges. No challenges can be added afterwards:

`func NewTranscript(h HashFS, challenges ...string) Transcript `

Transcript object offers the following API:
* `(m *Transcript) Bind(challenge string, value []byte) error`: binds a challenge (referred to by its name) to a value. An error is returned when the challenge has already been computed or if the name is not recorded
* `(m *Transcript) ComputeChallenge(challenge string) ([]byte, error)`: computes the challenge linked to the name. It returns an error if the previous challenge (in the order defined during the Transcript creation) is not computed.


fiat-shamir/transcript_test.go
```diff
@@ -0,0 +1,136 @@
...
+	fs.Bind("alpha", values[1])
+	fs.Bind("beta", values[2])
+	fs.Bind("beta", values[3])
+	fs.Bind("gamma", values[4])
+	fs.Bind("gamma", values[5])
+
+	return fs
+}
+
+func eq(a, b []byte) bool {
```
gbotrel, line 34: use `bytes.Equal(...)` here

polynomial/commitment.go
```diff
@@ -32,6 +32,7 @@ type Polynomial interface {
 type Digest interface {
 	io.WriterTo
```
gbotrel, line 33: are these two interfaces used? (io.WriterTo, io.ReaderFrom) on Digest?

fiat-shamir/transcript.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return res
+}
+
+// Bind binds the challenge to value. A challenge can be binded to an
+// arbitrary number of values, but the order in which the binded values
+// are added is important. Once a challenge is computed, it cannot be
+// binded to other values.
+func (m *Transcript) Bind(challenge string, value []byte) error {
+
```
gbotrel, line 113: cleaner to deal with error first ("idiomatic go"):

```
challengeNumber, ok := m.challengeOrder[challenge]
if !ok {
return err
}
// deal with nominal case. Same remark in ComputeChallenge
```

fiat-shamir/transcript.go
```diff
@@ -0,0 +1,171 @@
...
+			return m.challenges[challengeNumber], nil
+		}
+
+		m.h.Reset()
+
+		// write the challenge name, the purpose is to have a domain separator
+		bName := []byte(challenge)
+		_, err := m.h.Write(bName)
+		if err != nil {
+			return nil, nil
```
gbotrel, line 143: return nil, err ?

fiat-shamir/transcript.go
```diff
@@ -0,0 +1,171 @@
...
+		if challengeNumber != 0 {
+			if !m.isComputed[challengeNumber-1] {
+				return nil, errPreviousChallengeNotComputed
+			}
+			bPreviousChallenge := m.challenges[challengeNumber-1]
+			m.h.Write(bPreviousChallenge[:])
+		}
+
+		// write the binded values in the order they were added
+		m.h.Write(m.bindings[challengeNumber])
```
gbotrel, line 156: handle returned error

polynomial/commitment.go
ThomasPiellard: Currently there are not used, but they will be eventually (the commitments need to be passed to the rollup operator/blockchain at some point).

PR #45, opened 2021-04-27
Added optional precompute parameter in FFT domain initialization

### LINKED ISSUE #36: fft: remove extra bitReverse operations
Previous version of the `fft` included bit reverse operation in the domain pre-computed coset tables (in a dirty way, non generalizable to many cosets). 

 Need to propose a general pattern to avoid bitReverse operations.

ThomasPiellard:
This PR fixes #36  a performance loss of the FFT which occurred when the support of arbitrary cosets was added.

## Description

When creating an FFT domain, a boolean value is now passed as parameter to force some pre computations, namely the BitReverse of the coset table. Forcing the pre computing or not is a time/memory trade off. Forcing the pre computations doubles the memory needed to stored a domain.

## API breaking changes

`func NewDomain(m, depth uint64, precomputeReversedTable bool) *Domain`
instead of:
`func NewDomain(m, depth uint64) *Domain`

## Benchmarks

MBP 2.2GHz 6-core i7 - 16GB RAM

```
BenchmarkFFT/fft_2**18bits_(cosets_without_precomputations)-12                48          20933738 ns/op
BenchmarkFFT/fft_2**18bits_(cosets_with_precomputations)-12                   96          12964353 ns/op
```


gbotrel:
Looks good to me, merging. May revisit with optimization work on github.com/ConsenSys/gnark/issues/69

PR #112, opened 2021-12-14
Permit code to be compiled as a plugin with Go 1.17.x

ale-linux:
In Go version 1.17.x, `R15` is reserved for the GOT (see
https://github.com/golang/go/issues/48468) when the code is built as a
plugin, and code that uses `R15` will fail to build. This PR changes
the designation of registers to fix the compilation errors.

Signed-off-by: Alessandro Sorniotti <aso@zurich.ibm.com>

### DIFF
--- a/ecc/bn254/fp/element_mul_amd64.s
+++ b/ecc/bn254/fp/element_mul_amd64.s
@@ -75,7 +75,7 @@ TEXT ·mul(SB), $24-24
 
 	// A -> BP
 	// t[0] -> R14
-	// t[1] -> R15
+	// t[1] -> R12
 	// t[2] -> CX
 	// t[3] -> BX
 	// clear the flags
@@ -83,11 +83,11 @@ TEXT ·mul(SB), $24-24
 	MOVQ 0(R11), DX
 
 	// (A,t[0])  := x[0]*y[0] + A
-	MULXQ DI, R14, R15
+	MULXQ DI, R14, R12
 
 	// (A,t[1])  := x[1]*y[0] + A
 	MULXQ R8, AX, CX
-	ADOXQ AX, R15
+	ADOXQ AX, R12
 
 	// (A,t[2])  := x[2]*y[0] + A
 	MULXQ R9, AX, BX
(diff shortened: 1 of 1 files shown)


ale-linux:
See https://github.com/hyperledger/fabric/issues/3073 for the point where the build is broken

ecc/bn254/fp/element_mul_amd64.s
```diff
@@ -313,12 +313,12 @@ TEXT ·mul(SB), $24-24
 	ADCXQ AX, BX
 	ADOXQ BP, BX
 
-	// reduce element(R14,R15,CX,BX) using temp registers (R13,SI,R12,R11)
-	REDUCE(R14,R15,CX,BX,R13,SI,R12,R11)
+	// reduce element(R14,R12,CX,BX) using temp registers (R13,SI,R13,R11)
+	REDUCE(R14,R12,CX,BX,R13,SI,AX,R11)
```
denyeart, line 317: I followed up to here.
R15 becomes R12.
R12 becomes R13.
But why AX here?

ale-linux: It's indeed a typo - `AX` is used as a temp register by `REDUCE`. Will fix - thx Dave

gbotrel:
Interesting; it seems this happens when user assembly uses R15 and global variables in the same code. While your patch works for BN254 (thanks for the effort :-) ), it won't generalize well to the other curves supported in gnark-crypto, since they use all available registers. 

Found this PR that deals with the issue (and keep using R15) by moving global variables in defines: https://go-review.googlesource.com/c/go/+/36414/ . 

Will have a look and experiment with our x86 generator to make that work 👍 .

gbotrel:
( also, unrelated, I was not aware gnark-crypto was a direct dependency of hyperledger/fabric, to what extend is it used? )

ale-linux:
The approach detailed in that link sounds promising - though it might introduce a perf penalty. Definitely worth investigating though.

Fabric depends on gnark-crypto since it (fabric) pulls https://github.com/IBM/idemix/ which depends on https://github.com/IBM/mathlib/ which in turn uses gnark-crypto for one of the curves.

In that sense, my current fix would be an acceptable stop-gap solution till you find one that generalises to all curves/functions.

Obviously up to you/the maintainers of gnark-crypto to decide whether it's okay to stage changes this way.

gbotrel:
a fix has been merged to develop: https://github.com/ConsenSys/gnark-crypto/pull/114 . Essentially reproducing your solution, but in x86 code generator, and for another method too 👍 . 

should fix your issue on bn254. 

We are few days away from a new tag (v0.6.0) but you can safely use head on develop (f4bce86577924bace81a8f509584c38614df96dc).

ale-linux:
Many thanks, that was quick!

Issue #122, opened 2022-01-02
Faster subgroup checks on G2 on most curves


mratsim:
I've had a look in the library and it seems like Scott optimization in was implemented only for BLS12-381
(paper: https://eprint.iacr.org/2021/1130.pdf)

It is however applicable for all curves where gcd(h₁, h₂) == 1 with h₁ and h₂ the cofactors on G₁ and G₂.
In that case it is enough to check that `ψ(P) = [t-1]P` with t the trace.

As h₁ = 1 for BN curves and t-1 = 6u², it is applicable to BN254.
It is noted in the discussion that it also works for BLS24 curves.

yelhousni:
Hi @mratsim, I think that the G2 trick is implemented for [BLS12-381](https://github.com/ConsenSys/gnark-crypto/blob/d3277c3191f374355a2e3e3edd314552c4fd4ce3/ecc/bls12-381/g2.go#L376), [BLS12-377](https://github.com/ConsenSys/gnark-crypto/blob/d3277c3191f374355a2e3e3edd314552c4fd4ce3/ecc/bls12-377/g2.go#L376) and [BN254](https://github.com/ConsenSys/gnark-crypto/blob/d3277c3191f374355a2e3e3edd314552c4fd4ce3/ecc/bn254/g2.go#L376) (but not BLS24-315 indeed).

mratsim:
Oh indeed, I have a clone older than https://github.com/ConsenSys/gnark-crypto/pull/95/files

yelhousni:
So there was an error in Scott preprint for G2 membership test proof. However, the result is still correct (more on that here: https://eprint.iacr.org/2022/352.pdf).

PR #123, opened 2022-01-03
perf(bls24-315): faster G2 membership test

yelhousni:
#122

### DIFF
--- a/ecc/bls12-377/g1_test.go
+++ b/ecc/bls12-377/g1_test.go
@@ -124,6 +124,17 @@ func TestG1AffineIsOnCurve(t *testing.T) {
 		GenFp(),
 	))
 
+	properties.Property("[BLS12-377] IsInSubGroup and MulBy subgroup order should be the same", prop.ForAll(
+		func(a fp.Element) bool {
+			var op1, op2 G1Jac
+			op1 = fuzzJacobianG1Affine(&g1Gen, a)
+			_r := fr.Modulus()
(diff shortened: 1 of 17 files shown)


ThomasPiellard:
@yelhousni Can you add a specific test for `IsInSubGroup` ?

Also there is something that bothers me in [https://eprint.iacr.org/2021/1130.pdf](https://eprint.iacr.org/2021/1130.pdf)... The reasoning that I understood from the paper is "psi acts as the multiplication by u on the p+1-t torsion of E, so let's just check that psi(Q)=[u]Q".

 However, nothing says that there is no other torsion subgroup on which psi acts like this. For example, if we find `a` such that 'a | h2', `p^2-1 != 0 [a]` (so E' contains a subgroup of order `a` but not the full `a`-torsion) and `u` is a root of `X^2-tX+p [a]`, then psi acts by multiplication by `u` on the `a` torsion of E'[Fp^2]. From the equations, it's not clear that such a situation does not happen. It seems really unlikely because of the third condition (that `u` has to be a root of `X^2-tX+p [a]`), but unless proceeding case by case there is no guarantee of that.

So I am in favour of keeping the solution before the very end of sect 4, that is checking if psi(uQ)+psi(Q)+..=uQ. The delta that I have between the benches is 
```
benchmark                         old ns/op     new ns/op     delta
BenchmarkG2JacIsInSubGroup-12     122793        126198        +2.77%
```

ThomasPiellard (approved):

PR #129, opened 2022-01-19
GT compression

yelhousni:
Compress `F_{p^k}` elements (E24, E12, E6) when they are in the cyclotomic subgroup, i.e. `z^{\Phi_k(p)} == 1`. This works in particular to compress GT elements after a pairing. The compression ratio is 1/2 using Torus-based T2 lossless compression following ["COMPRESSION IN FINITE FIELDS AND TORUS-BASED CRYPTOGRAPHY", K. RUBIN AND A. SILVERBERG](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.90.8087&rep=rep1&type=pdf) (section 2). 

Note we can have a lossless compression with ratio 1/3 using CEILIDH (section 3) if we decide to change the extensions tower (e.g. Fp2-Fp4-Fp12 for E12). This would result is a slightly slower pairing though.


mratsim:
See also Benger-Scott paper https://eprint.iacr.org/2009/556.pdf

image

Devigili et al https://eprint.iacr.org/2006/471.pdf has the costs between the different towerings and direct sextic:
image
image
image

Timings

image

It seems like the same perf.

And we can change representation (https://github.com/supranational/blst/issues/101#issuecomment-1015511522, https://github.com/supranational/blst/pull/102#pullrequestreview-883682769)

gbotrel (approved):

ThomasPiellard (approved):

PR #130, opened 2022-01-19
Sswu for bls12-381 G1

Tabaie:
Needs a lot of cleanup. Draft PR to discuss code structure.

Main TODOs:

- Move hash pre-computations out of main config class (used in runtime)
- Hard-code all test vectors
- Test sqrtRatio for many Fp moduli


Tabaie:
TODOs in previous comment done.

As is, gives EncodeToCurve and HashToCurve for bls12-381 G1. The code is generic and can be used for any elliptic curve over Fp given certain suite parameters ( a field element Z, an isogeny if necessary).

Next steps (next PR):

1. Extend the methods for Fp towers (so that for example we can have the methods for bls12-381 G2 as well)
2. Explore if suggested suites are emerging in the community for other curves. If not, make our own?

internal/generator/ecc/template/tests/sswu-fp.go.tmpl
```diff
@@ -0,0 +1,45 @@
...
+    "github.com/consensys/gnark-crypto/ecc/{{.Name}}/fp"
+	{{- if or (eq .CoordType "fptower.E2") (eq .CoordType "fptower.E4") }}
+	"github.com/consensys/gnark-crypto/ecc/{{.Name}}/internal/fptower"
+	{{else}}
+	"github.com/consensys/gnark-crypto/ecc/{{.Name}}/fp"
+	{{- end}}
+	"testing"
+)
+
+func TestSqrtRatio(t *testing.T) {
```
gbotrel, line 11: can use `gopter` property based testing as in other tests

internal/generator/ecc/template/sswu-fp.go.tmpl
```diff
@@ -0,0 +1,397 @@
...
+
+{{ if $g1Isogeny }}
+	//TODO: Add in E' first, then apply isogeny
+	isogenyG1(&Q0)
+	isogenyG1(&Q1)
+{{ end }}
+
+	var _Q0, _Q1, _res G1Jac
+	_Q0.FromAffine(&Q0)
+	_Q1.FromAffine(&Q1)
```
gbotrel, line 390: could remove either `_res` and `_Q1` ?

internal/generator/ecc/template/sswu-fp.go.tmpl
```diff
@@ -0,0 +1,397 @@
...
+
+	var tv3 {{.CoordType}}
+	//Standard doc line 5
+	var tv4 {{.CoordType}}
+	tv4.SetOne()
+	tv3.Add(&tv2, &tv4)
+	tv3.Mul(&tv3, &{{.CoordType}}{ {{ words64 .B }} })
+
+	tv4 = {{.CoordType}}{ {{ words64 .A }} }
+	//TODO: Std doc uses conditional move. If-then-else good enough here?
```
gbotrel, line 295: side remark; if we want constant time, in many cases, if / else won't be enough if the branch execution time are different --> 
it is cleaner to execute both branches, and then have a function that conditionally move the result in another variable. (assembly with CMOV on x86).

internal/generator/ecc/template/sswu-fp.go.tmpl
```diff
@@ -0,0 +1,397 @@
...
+
+   *z = tv3
+   return isQr
+}
+{{ end }}
+
+//TODO: Use addchain
+//TODO: Might duplicate functionality from mulByConst functions
+// mulByZ multiplies x by {{.Z}} and stores the result in z
+func mulByZ(z *{{.CoordType}}, x *{{.CoordType}}) {
```
gbotrel, line 246: unless Z is often small on our curves, I suspect the perf diff with the optimized Mul operation are negligible . 
I would kill this func and use Mul directly.

internal/generator/ecc/template/sswu-fp.go.tmpl
```diff
@@ -0,0 +1,397 @@
...
+// sqrtRatio computes the square root of u/v and returns true if u/v was indeed a quadratic residue
+// if not, we get sqrt(Z * u / v). Recall that Z is non-residue
+// The return value is undefined for u = 0
+// Taken from https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/13/ F.2.1.1. for any field
+// The main idea is that since the computation of the square root involves taking large powers of u/v, the inversion of v can be avoided
+func sqrtRatio(z *{{.CoordType}}, u *{{.CoordType}}, v *{{.CoordType}}) bool {
+   tv1 := {{.CoordType}} { {{words64 (index .SqrtRatioParams 5) }} }
+
+   var exp big.Int
+   exp.SetBytes([]byte { {{ printList (bytes (index .SqrtRatioParams 3) ) }} })
```
gbotrel, line 190: big.Int SetBytes is not very readable on the generated code, why not set from a hex string value?

internal/generator/ecc/template/sswu-fp.go.tmpl
gbotrel: additional precisions; IF / ELSE are, to my knowledge, most of the time translated by JMP instructions in assembly.

internal/generator/config/curve.go
```diff
@@ -69,3 +94,178 @@ func newFieldInfo(modulus string) Field {
...
+		i--
+
+		res.Mul(res, res).Mod(res, mod)
+	}
+
+	res.Mod(res, mod)
+	return res
+}
+
+func pow(p *big.Int, pow uint8) *big.Int {
```
gbotrel, line 201: why not use big.Int Exp method here? for powMod I understand big.Int Exp followed by a mod is going to be terribly inefficient. but simple pow?

internal/generator/config/curve.go
```diff
@@ -69,3 +94,178 @@ func newFieldInfo(modulus string) Field {
 	F.Modulus = func() *big.Int { return new(big.Int).Set(&bModulus) }
 	return F
 }
+
+// twoFactor returns the greatest m such that 2ᵐ divides x-1
+func twoFactor(x *big.Int) int {
+	if x.BitLen() <= 1 {
+		return -1 //substitute for ∞
+	}
+	if x.Bit(0) == 0 {
```
gbotrel, line 103: see big.Int `TrailingZeroBits` method to simplify this code

internal/generator/config/curve.go
```diff
@@ -22,6 +23,29 @@ type Curve struct {
...
+	DenHex []string //Denominator is monic. The leading coefficient (1) is omitted.
+}
+
+type HashSuite struct {
+	AHex string
+	BHex string
+
+	Z int // z (or zeta) is a quadratic non-residue with //TODO: some extra nice properties, refer to WB19
+
+	Isogeny *Isogeny //pointer so it's nullable. TODO: Bad practice or ok?
```
gbotrel, line 48: common practice (have pointers to have a "zero value" set to the nil pointer)

internal/generator/config/curve.go
```diff
@@ -22,6 +23,29 @@ type Curve struct {
...
+
+type Isogeny struct {
+
+	//Isogeny to original curve
+	XMap RationalPolynomial
+	YMap RationalPolynomial // The y map is also evaluated on x. The result is multiplied by y.
+}
+
+type RationalPolynomial struct {
+	NumHex []string
```
gbotrel, line 38: naming wise, I would remove the "Hex" suffix. just a comment that says coefficients are stored in base16 strings is OK.

internal/generator/config/curve.go
```diff
@@ -1,11 +1,12 @@
 package config
 
 import (
-	"math/big"
-
 	"github.com/consensys/gnark-crypto/field"
+	"math/big"
 )
 
+//TODO: Same data structure used in runtime and for code generation. HashInfo not needed for runtime
```
gbotrel, line 8: yes, we should add (separate PR / ticket) build tags for the code generation ,to ensure we don't pollute API or binary with useless structs needed only for code gen.

field/internal/templates/element/temp_for_hash.go
```diff
@@ -0,0 +1,39 @@
...
+// Taken from https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/ section 4.1
+// The sign of an element is not obviously related to that of its Montgomery form
+func (z *{{.ElementName}}) Sgn0() bool {
+	nonMont := *z
+	nonMont.FromMont()
+
+	return nonMont[0]%2 == 1
+}
+
+func (z *{{.ElementName}}) SetHex(hex string) {
```
gbotrel, line 16: remove this function, Element.SetString() works with both hex (prefixed by 0x) and decimal strings, no need to add this extra API.

field/internal/templates/element/temp_for_hash.go
```diff
@@ -0,0 +1,39 @@
...
+	var i big.Int
+	i.SetString(hex, 16)
+	if _, b := i.SetString(hex, 16); !b {
+		panic("SetString failed")
+	}
+	z.SetBigInt(&i)
+}
+
+func (z *{{.ElementName}}) EvalPolynomial(monic bool, coefficients []{{.ElementName}}, x *{{.ElementName}}) {
```
gbotrel, line 25: monic boolean a part, this is duplicate with `polynomial` package. 
This API shouldn't live in the field Element package; see with @ThomasPiellard

field/internal/templates/element/ops_generic.go
```diff
@@ -7,21 +7,18 @@ const OpsNoAsm = `
 // or side-channel attack resistance
 // /!\ WARNING /!\
 
+// TODO: Discuss why this has to be in place, and why x is not a receiver
```
gbotrel, line 10: --> x not receiver --> assembly implementation constraints / perf.

field/internal/templates/element/ops.go
```diff
@@ -5,6 +5,8 @@ const OpsAMD64 = `
 
 {{if .ASM}}
 
+//TODO: Add MulBy11
```
gbotrel, line 8: MulBy11 may not be that much faster than a Mul.

field/internal/templates/element/inverse.go
```diff
@@ -58,10 +58,6 @@ const invIterationsN = {{.P20InversionNbIterations}}
 // Implements "Optimized Binary GCD for Modular Inversion"
 // https://github.com/pornin/bingcd/blob/main/doc/bingcd.pdf
 func (z *{{.ElementName}}) Inverse(x *{{.ElementName}}) *{{.ElementName}} {
-	if x.IsZero() {
```
gbotrel, line 61: should be in another PR

field/internal/templates/element/base.go
```diff
@@ -208,6 +208,11 @@ func (z *{{.ElementName}}) IsZero() bool {
 	return ( {{- range $i :=  reverse .NbWordsIndexesNoZero}} z[{{$i}}] | {{end}}z[0]) == 0
 }
 
+// IsOne returns z == 1 in Montgomery
```
gbotrel, line 211: remove Montgomery from comment 👍

field/field.go
```diff
@@ -303,3 +302,20 @@ func extendedEuclideanAlgo(r, q, rInv, qInv *big.Int) {
...
+
+	var i big.Int
+	i.SetString(hex, 16)
+	IntToMont(&i, fieldModulus)
+
+	return i
+}
+
+func IntToMont(i *big.Int, fieldModulus *big.Int) {
+	nbWords := (fieldModulus.BitLen()-1)/64 + 1
```
gbotrel, line 318: does this work on 32bit architecture? 

Can we define these methods on the Field object to ensure we use the same R?

gbotrel (changes requested):

internal/generator/config/curve.go
Tabaie: Not sure I understood. So this _is_ common practice?

gbotrel: yes

internal/generator/config/curve.go
Tabaie: I got the impression from the API that only modular exponentiation is supported. It seems like setting the modulus to nil gives integer exp.

internal/generator/ecc/template/sswu-fp.go.tmpl
Tabaie: It _is_ small, but we can't use the same trick (or a simple-enough-to-be-worth-it variant) on extensions.

internal/generator/ecc/template/sswu-fp.go.tmpl
Tabaie: Reading from hex string is very slow (20% of entire runtime initially). But I can add a comment for readability.

field/internal/templates/element/temp_for_hash.go
Tabaie: So just remove the monic (over-)optimization and use `polynomial`? Because as I remember that was the only reason I didn't use it.

gbotrel: Yes 👍  but double check with @ThomasPiellard plan to kill the `polynomial` package. I think "Eval" method should exist somewhere.

ThomasPiellard: yes, so for this polynomial thing, either we kill the package and put Eval somewhere (in a utils file perhaps), or we can keep the package only for aliasing a Polynomial type which would be a []fr.Element, and just keep the Eval method on it.

Tabaie: I'm good with both. In the meantime should I use the polynomial package?

gbotrel: That is a very good candidate for generic 🙄

ecc/bls12-381/sswu-fp.go
```diff
@@ -0,0 +1,314 @@
...
+	var y1 fp.Element
+	gx1Square := sqrtRatio(&y1, &tv2, &tv6)
+
+	var y fp.Element
+	y.Mul(&tv1, u)
+
+	//Standards doc line 20
+	y.Mul(&y, &y1)
+
+	//TODO: Not constant time. Is it okay?
```
yelhousni, line 244: ideally we would want a constant-time algorithm (but many algorithms in gnark-crypto are not)

ecc/bls12-381/sswu-fp.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	//TODO: Not constant time
+	if sgn0(u) != sgn0(&y) {
+		y.Neg(&y)
+	}
+
+	//Standards doc line 25
+	//TODO: Not constant time. Use Jacobian?
+	x.Div(&x, &tv4)
```
yelhousni, line 257: slow Div.. how does the algorithm perform when implemented fully in jacobians?

ecc/bls12-381/sswu-fp.go
```diff
@@ -0,0 +1,314 @@
...
+	if isQr {
+		*z = y1
+	} else {
+		*z = y2
+	}
+
+	return isQr
+}
+
+//TODO: Use addchain
```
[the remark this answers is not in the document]
Tabaie: I meant the second, and what you say makes sense. In any case, @gbotrel is suggesting that in the long run (with field extensions) we will probably end up doing a regular `Element.Mul` anyway.

ecc/bls12-381/sswu-fp.go
Tabaie: Conditional move is in the works!

ecc/bls12-381/sswu-fp.go
yelhousni: #148

yelhousni (approved):
LGTM

PR #143, opened 2022-01-31
Feat/cmov

Tabaie:
`Select` function with generic as well as x86 assembly implementations.


gbotrel:
```
func BenchmarkElementSelect(b *testing.B) {
	var x, y Element
	x.SetRandom()
	y.SetRandom()

	b.Run("generic", func(b *testing.B) {
		b.ResetTimer()

		for i := 0; i < b.N; i++ {
			_selectGeneric(&benchResElement, int64(i%3), &x, &y)
		}
	})

	b.Run("asm", func(b *testing.B) {
		b.ResetTimer()

		for i := 0; i < b.N; i++ {
			select_(&benchResElement, int64(i%3), &x, &y)
		}
	})

}

```

gives the `generic` version faster than the asm one on my amd desktop for all moduli we have except the fat ones from bw6-xxx, where x86 asm is slightly faster. 

No point in having an assembly version for this func, the generic one is enough 👍

gbotrel (changes requested):
remove x86 impl, keep only generic 👍

Tabaie:
Done!

PR #144, opened 2022-02-02
feat: constant time Equal with tests and benchmarks

Tabaie:
Slightly faster than `Equals` on small moduli, up to %50 slower on big ones
May be useful in algorithms advertised as c-time like sswu hash


gbotrel:
lgtm 👍 
But, given that perf is so close, I'ld avoid surcharging the `Element` API; replace the old "Equal" with the new one (and no need to decompose in 2 methods with Diff / EqualCT). 

@yelhousni I did a quick search and I don't think equal is perf critical anywhere in the pairing or msm, is it? Impact of this PR --> for BW6 curves (not for the others) the equal when field elements are different becomes twice slower (we're talking 2ns to 4ns... )

yelhousni:
I think it's fine to replace the old `Equal` with the new c-time one. We would rather get a c-time hash-to-curve than a slower pairing check on BW6 curves (which occurs only when there is a problem).

PR #145, opened 2022-02-03
Fix/fft cosets

ThomasPiellard:
This PR modifes the FFT so the cosets are taken in Fr* \ (Z/2^nZ) where (Z/2^nZ) is identified with the maximum power-of-two multiplicative subgroup of Fr*, instead of being taken in in (Z/2^nZ) which prevented us from using the full two-adicity.

Plookup, permutation argument, etc have been modified accordingly.

API breaking changes:
* `NewDomain(m, depth uint64, precomputeReversedTable bool)` -> `NewDomain(m)`
*  `FFT(a []fr.Element, decimation Decimation, coset uint64)` ->  `FFT(a []fr.Element, decimation Decimation, coset ...bool)`. When coset is specified and set to `true`, the FFT is done on a coset (precomputed and not up to the user).

It fixes #145 .


gbotrel (changes requested):
couple of addchain old binary files to remove from this PR, must be a reliq from a old version + git add.

PR #153, opened 2022-02-22
Remove some useless code in fft.NewDomain

zhiqiangxu:
Just stumpled upon some useless code.

### DIFF
--- a/ecc/bls12-377/fr/fft/domain.go
+++ b/ecc/bls12-377/fr/fft/domain.go
@@ -75,7 +75,6 @@ func NewDomain(m uint64) *Domain {
 
 	rootOfUnity.SetString("8065159656716812877374967518403273466521432693661810619979959746626482506078")
 	const maxOrderRoot uint64 = 47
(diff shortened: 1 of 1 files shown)


ThomasPiellard:
Hi @zhiqiangxu , yes that's right thanks for spotting this. There is also `PrecomputeReversedTable` which is not used. Could you modify your PR to merge it to develop instead of master?

ThomasPiellard:
Yes I will remove it, some comments need to be updated as well. Normally if I am not mistaken after the removal `PrecomputeReversedTable` only the marshal functions will be affected...

zhiqiangxu:
> Yes I will remove it, some comments need to be updated as well. Normally if I am not mistaken after the removal of `PrecomputeReversedTable` only the marshal functions will be affected...

Right.

zhiqiangxu:
> I didn't pay attention but in fact you didn't modify the templates for the code generation so there is some additional code to write. To save your commit, could you tweak the PR again to merge it to `feat/fft_cleanup` (I juste created this branch). I can add the other required modifications right now.

Done!

Issue #155, opened 2022-03-11
Merge SW (`hash_to_curve.go`) and SSWU (`sswu_g*.go`)


yelhousni:
Currently we support SW and SSWU hash-to-curve maps. We should only keep SSWU. However, when the isogeny degree of some curves is big we migt want to keep SW for this case instead. Also both should be in the same file.

yelhousni:
It's more of a cleanup task so low to medium priority but it can be done quickly. We just need to generify SW hash-to-curve for the remaining cases: BN254 (G1 and G2) and get rid of the `hash_to_curve.go`.

yelhousni:
Something that is high priority and related to this is SSWU for BLS24 G2 (Fp4) #162 .

PR #156, opened 2022-03-11
SSWU parameters

yelhousni:
This PR adds parameters for SSWU:
- [x] On G1: BLS12-377, BLS24-315, BW6-761 and BW6-633 with a 2-isogeny 
- [x] On G2: BW6-633 with a 5-isogeny

Other curves, on G2:
- [ ] BN254 with a 193-isogeny (too big we might want to stick to generic SW)
- [ ] BLS24-315 with a 5-isogeny (`Fp4` WIP)
- [x] BLS12-377 with a *23*-isogeny <s>85752112189-isogeny</s> on G2 <s>(too big we might want to stick to generic SW)</s>
- [x] BW6-761 with a 37-isogeny (too big? we might want to stick to generic SW)

on G1:
- [ ] BN254 with a 59-isogeny (too big we might want to stick to generic SW)

This will be addressed in a different PR. Also see #155 .


Tabaie:
A 39-isogeny may still be good, as I remember the 11-isogeny in BLS12-381 took a very small portion of runtime. Should we generify the SW implementation from BLS12-381 to use for BLS12-377?

yelhousni:
> A 39-isogeny may still be good, as I remember the 11-isogeny in BLS12-381 took a very small portion of runtime. Should we generify the SW implementation from BLS12-381 to use for BLS12-377?

Yes we might give the 39-isogeny SSWU a try and yes we should generify th SW map for the remaining cases. We can take the implementation from any curve as the code is the same (`b=0`).

yelhousni:
Last commit adds parameters of a 37-isogeny for the G2 bw6-761 case. I think a degree 37 is still fine.  G1 and G2 for BW6 are both on `Fp` so comparing G1 SSWU with a 2-isogeny to G2 SSWU with a 37-isogeny is `~35 us` difference.

yelhousni:
and I was mistaken there is a suitable 23-isogeny on G2 for BLS12-377 (commit fadf1ae193903b95df2725f399bb4930f315d94e). So we should get rid of the slower SW map and keep only SSWU (#155).

yelhousni:
#158

PR #161, opened 2022-03-17
Feat/inverse verification

Tabaie:
The condition for this to work is slightly weaker than NoCarry but I think it uses a couple of multiplication functions that use NoCarry-like optimizations (mulW).


gbotrel:
Something bothers me; why do we ignore the carry here, in the `linearCombNonModular` function?: 

```
	yHi, _ = bits.Add64(xHi, yHi, carry)
```

Effectively, this function does : 
> linearCombNonModular sets z = (x * xC) + (y * yC) without modular reduction; 

First calls looks good since `updateFactorsDecompose` produces outputs who are small; 

```
f1, c1 = updateFactorsDecompose(c1)
bHi := b.linearCombNonModular(&s, f1, &b, c1)
``` 

But this call: 

```
// c₀ aliases f₀, c₁ aliases g₁
			c0, g0, f1, c1 = c0*pf0+g0*pf1,
				c0*pg0+g0*pg1,
				f1*pf0+c1*pf1,
				f1*pg0+c1*pg1

			s = u

			// 0 ≤ u, v < 2²⁵⁵
			// |F₀|, |G₀| < 2⁶³
			u.linearComb(&u, c0, &v, g0)
```

The condition we have is `|F₀|, |G₀| < 2⁶³` and sure enough, if you test this function with c0, g0 == - 2^63 - 1 , there is a carry.

Adding this; 

```
yHi, carry = bits.Add64(xHi, yHi, carry)

	if carry != 0 {
		panic("here")
	}
```

Crash the fuzzing test; (but the inverse result is otherwise correct). Hence my question; why do we ignore the carry here?

PR #173, opened 2022-03-30
Perf/arm assembly: The "easy stuff"

Tabaie:
add, sub, double, neg


gbotrel:
@Tabaie what's the status on that? Can't test locally since a dep (bavard) is pointing to a local version

Tabaie:
Only two out of four (add and neg I think) offer speedup as of now, so I'm not sure if it's worth complicating the code base. The changes to bavard are related to ARM code generation which could be worth merging in any case.

PR #176, opened 2022-04-25
Add a curve defined on 1-limb

yelhousni:
This PR adds a full implementation of a pairing-friendly BLS12 curve defined over a field of 39 bits and has a subgroup of order of 27 bits.

(Maybe we should rename it to `toy_bls12` to avoid confusion)

Edit: to merge after #177  and #178


gbotrel:
closing this, but keeping the branch in case 👍

PR #178, opened 2022-04-30
field arithmetic cleanup

gbotrel:
- test: added `BatchInvert` test
- style: remove global variable for `qMinusTwo`
- fix: handle negative exponent for `field.Exp`


yelhousni (approved):
This sounds correct, we can merge 👍
Not sure we need `BatchExp()` but it could benefit from `BatchInvert()`

Issue #163, opened 2022-03-22
Faster multiplication in extension fields


yelhousni:
https://eprint.iacr.org/2022/367 by Patrick Longa.

The author says it results in a ~X1.3 faster pairing on BLS12-381. We might use this work for all extension fields we have so far. 

There is already an implementation (not the pairing one) integrated in SIDH (https://github.com/microsoft/PQCrypto-SIDH).

yelhousni:
It seems to me that this would not be compatible with the optimised field multiplication we have (https://hackmd.io/@gnark/modular_multiplication). So closing this.

Issue #79, opened 2021-10-14
Quadruple-and-add for BLS24 Miller loop


yelhousni:
Similarly to BLS12 curves, BLS24 Miller loop counter is the curve seed `u`. Although this seed is 32 bits (half the size of BLS12 seeds), the BLS24 Miller loop is slower. This is mainly because the most costly operations encountered in the Miller loop computation are those that take place in the full extension field `Fp24`. Implementing this paper's [[CBNW10]](https://eprint.iacr.org/2010/104.pdf) quadruple-and-add algorithm would reduce operations in `Fp24` (see section 6 for comparisons).

Raneet10:
Hi! Has this been or being worked on ? In case it hasn't been (or isn't being), I would like to take a stab at it, given that it's a good first issue. Thanks!

yelhousni:
> Hi! Has this been or being worked on ? In case it hasn't been (or isn't being), I would like to take a stab at it, given that it's a good first issue. Thanks!

Hi @Raneet10 ! No this is still an open issue — go for it! It might become handy in the near future.

yelhousni:
BLS24 is low priority today so closing this.

PR #179, opened 2022-05-04
Add BLS24-317 curve

yelhousni:
Optimized curve for KZG:
- `Fp2[u]=Fp/x^2+1`
- `2^60 | r-1`
- `HW2NAF(x0)=6`
- security: 127 bits


yelhousni:
Compared to BLS12-381, the commitment and opening are faster (but the verification is slower because of the `Fp24` arithmetic - #163 and #79 might help though):
```
benchmark                        bls12-381 ns/op     bls24-315 ns/op     delta
BenchmarkKZGCommit-32            30541552            23679376           -22.47%
BenchmarkKZGOpen-32              32462285            25574709           -21.22%
BenchmarkKZGBatchOpen10-32       40809728            33966791           -16.77%
BenchmarkKZGVerify-32            1419455             3550663            +150.14%
BenchmarkKZGBatchVerify10-32     1806584             3903845            +116.09%
```

yelhousni:
Also some x86 asm acceleration for `Fp2` might help as it's the case for BLS24-315:
```
Benchmark                        bls12-381 ns/op     bls24-315 ns/op     delta
BenchmarkKZGVerify-32            1419455             2862937            +101.69%
BenchmarkKZGBatchVerify10-32     1806584             3193672            +76.78%
```

With this, one would expect the `delta` to be smaller for BLS24-317 because of `Fp2` non-residue and the seed low Hamming weight.

gbotrel (approved):

PR #183, opened 2022-05-10
refactor: field mod base 10 unless prefixed by 0x. 'Field.ModulusDec'

Tabaie:
I personally get confused sometimes about the naming of the field "modulus". Not always sure if it's decimal or allowed to be hex. Thought it would be beneficial to be explicit about the base.

Also being a bit more strict about input modulus formatting to `NewField`. Technically the string `"123"` could be taken to be in either base 10 or 16. I changed that to assume base 10 unless prefixed by `"0x"` or `"0X"`.

### DIFF
--- a/field/field.go
+++ b/field/field.go
@@ -18,11 +18,10 @@ package field
 import (
 	"errors"
 	"fmt"
+	"github.com/consensys/gnark-crypto/field/internal/addchain"
 	"math"
 	"math/big"
 	"math/bits"
-
-	"github.com/consensys/gnark-crypto/field/internal/addchain"
 )
 
 var (
@@ -34,7 +33,7 @@ type Field struct {
 	PackageName               string
 	ElementName               string
 	ModulusBig                *big.Int
-	Modulus                   string
+	ModulusDec                string
 	ModulusHex                string
 	NbWords                   int
 	NbBits                    int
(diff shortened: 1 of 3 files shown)


field/field.go
```diff
@@ -76,15 +75,22 @@ type Field struct {
 func NewField(packageName, elementName, modulus string, useAddChain bool) (*Field, error) {
 	// parse modulus
 	var bModulus big.Int
-	if _, ok := bModulus.SetString(modulus, 0); !ok {
+	var base int
```
gbotrel, line 78: OK for `Modulus` -> `ModulusHex`, but this part is redundant /  not useful `if modulus[0] == '0' && (modulus[1] == 'x' || modulus[1] == 'X') {` . 

I'ld rather copy paste the `big.Int` `SetString(x, 0)` doc (https://pkg.go.dev/math/big#Int.SetString ) :

> The base argument must be 0 or a value between 2 and MaxBase. For base 0, the number prefix determines the actual base: A prefix of “0b” or “0B” selects base 2, “0”, “0o” or “0O” selects base 8, and “0x” or “0X” selects base 16. Otherwise, the selected base is 10 and no prefix is accepted.

gbotrel (changes requested):

field/field.go
Tabaie: So all that's left is changing `Field.Modulus` to `Field.ModulusDec`. Should we keep that one?

gbotrel: I don't feel strongly about it -- since the "default" by convention in gnark-crypto is base10, then it's implicit. 

But it's internal code gen, if you strongly prefer it, I'm not going to fight you on that one :-)

Tabaie: > OK for Modulus -> ModulusHex,
I think you mean `Modulus` -> `ModulusDec`?

Tabaie: > I don't feel strongly about it -- since the "default" by convention in gnark-crypto is base10, then it's implicit.
But it's internal code gen, if you strongly prefer it, I'm not going to fight you on that one :-)

For the record, this is in response to a now deleted comment asking about whether @gbotrel is okay with the `Modulus`->`ModulusDec` change.

PR #188, opened 2022-05-16
Hash2Curve uniform interface

Tabaie:


Tabaie:
@yelhousni There are errors with bw6-633 G1 and bw6-761 G2. The combination of the SSWU and the isogeny is not on the curve. It could either be that the SSWU parameters are incorrect, or the isogeny. Would you take a look at the isogenies? In the meantime I'll develop a test for the map itself to see if the output is actually on E'.

Tabaie:
It seems that all the SSWU are "correct" (in the sense that the output is always on E')

yelhousni:
@Tabaie So there were some errors in ordering the isogeny denominator coefficients for bw6-761 G2 and a duplicate coefficient in bw6-633 G1. Now everything is corrected and tests pass.

gbotrel:
@Tabaie so API wise, we have 
* `HashToG1`
* `EncodeToG1`
* `MapToG1`

In the signatures, `(msg, dst []byte) (G1Affine, error) ` What is `dst` for ?  It is a bit confusing for an API to have different outputs both as params and returns, at least deserve a bit more doc 👍

yelhousni:
> In the signatures, `(msg, dst []byte) (G1Affine, error) ` What is `dst` for ? It is a bit confusing for an API to have different outputs both as params and returns, at least deserve a bit more doc 👍

`dst` is a domain separation tag to guarantee independent outputs in case of multiple hashes/encodings (see discussion [here](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-06#section-3.1)).

gbotrel:
I see I thought dst == destination

ecc/bls12-378/hash_to_g2.go
```diff
@@ -0,0 +1,146 @@
...
+	var _Q0, _Q1, _res G2Jac
+	_Q0.FromAffine(&Q0)
+	_Q1.FromAffine(&Q1)
+	_res.Set(&_Q1).AddAssign(&_Q0)
+	res.FromJacobian(&_res)
+	return res, nil
+}
+
+// returns false if u>-u when seen as a bigInt
+func sign0(u fp.Element) bool {
```
gbotrel, line 140: @Tabaie can we use fp.Element.LexicographicallyLargest here?

ecc/bls12-378/hash_to_g2.go
```diff
@@ -0,0 +1,146 @@
...
+// MapToG2 maps an fp.Element to a point on the curve using the Shallue and van de Woestijne map
+// https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-06#section-2.2.1
+func MapToG2(t fptower.E2) G2Affine {
+	res := svdwMapG2(t)
+	res.ClearCofactor(&res)
+	return res
+}
+
+// EncodeToG2 maps an fp.Element to a point on the curve using the Shallue and van de Woestijne map
+// https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-06#section-2.2.2
```
gbotrel, line 102: @Tabaie can we update the doc of the public functions (EncodeXXX, MapXX, ...) to explain the parameters? 
For example, DST is not clear if you don't jump on the specs. Some function says "map an fp.Element" but takes as param a []byte.

Tabaie:
@gbotrel dst is domain separation tag actually.

ecc/bls12-378/hash_to_g2.go
Tabaie: I think so.

gbotrel: can you investigate a bit? Are you sure it's coming from that specific commit? (I also don't obviously see why it happened, maybe it uncovered a bug somewhere we need to be aware of 👍 )

PR #204, opened 2022-06-01
feat: closes #137 moves consensys/goff into field/goff

### LINKED ISSUE #137: Move goff binary in field/goff
See https://github.com/ConsenSys/goff/issues/29 --> goff repo is not maintained and would benefit being attached to gnark-crypto.

gbotrel:
- refactor: move template field into internal/field
- feat: added field/Field interface and ecc.ID.ScalarField()
- feat: closes #137 moves consensys/goff into field/goff


internal/field/internal/templates/element/tests.go
```diff
@@ -10,7 +10,7 @@ import (
 	"math/bits"
 	"fmt"
 	{{if .UsingP20Inverse}} 
-	"github.com/consensys/gnark-crypto/field"
+	"github.com/consensys/gnark-crypto/internal/field"
```
omerfirmak, line 13: Because of this, generated files can not be embedded in another project as a standalone library.

PR #209, opened 2022-06-09
Unifying Hash To Curve code generation

Tabaie:
Unifying code generation for SvdW and SSWU


Tabaie:
I haven't added parameters for some of the other curves in there yet. Will do that after another PR simplifying SSWU code generation, removing generation-time parameter computation, after which I can delete all field arithmetic logic. Currently, if we try to develop field config classes for degree 4 extensions, we would have to implement arithmetic in towers as well which I would like to avoid.

yelhousni (approved):
LGMT! we can merge this.

Also the new svdw is 9% faster for BN254 G1 (and 4% for G2).

Tabaie:
Nice! It's probably because of how the precomputed values are loaded.

PR #217, opened 2022-06-30
feat: supsub

Tabaie:
Integrate the superscript and subscript functions into gnark-crypto (superscript is useful for pretty printing of polynomials)

### DIFF
--- a/internal/field/field.go
+++ b/internal/field/field.go
@@ -431,9 +431,7 @@ func CoordNameForExtensionDegree(degree uint8) string {
 }
 
 func (f *FieldConfig) WriteElement(element Element) string {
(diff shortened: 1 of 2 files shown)


gbotrel (approved):

gbotrel:
lgtm, maybe nitpicking; since you added that in `utils` package, some method will read like `utils.Sub(..)` which in the context of the rest of the library sounds a bit confusing :)

Tabaie:
> lgtm, maybe nitpicking; since you added that in `utils` package, some method will read like `utils.Sub(..)` which in the context of the rest of the library sounds a bit confusing :)

That's a great point. I'll change it to `ToSubscript`.

PR #220, opened 2022-07-13
[perf] KZG verify

yelhousni:
KZG verification algorithm consists in some scalar multiplications and then a pairing computation. The scalar multiplications are best implemented using Jacobian coordinates while the pairing is best implemented using mixed affine and projective coordinates. The `ScalarMul()` API makes unnecessary conversions between affine and Jacobians which costs some unnecessary inverses. Few solutions:
- [x] This PR implements a `ScalarMul()` functions that takes an affine point and returns a Jacobian.
- ~~[ ] Implementing the pairing with only projective coordinates and implementing the `ScalarMul()` in projective too (no conversions at all). This is implemented in `perf/kzg-verify-proj` branch but it is worth it only if the Hamming weight of the Miller loop size is **less than 6** (perf threshold for mixed add/dbl vs. proj add/dbl). This is not the case for the curve in gnark-crypto. Also it duplicates the pairing code.~~
- [ ] we can also implement a custom `BatchFromJacobian()` functions that takes elements from both G1 (Fp) and G2 (Fp2 or Fp4).

### DIFF
--- a/ecc/bls12-377/fr/kzg/kzg.go
+++ b/ecc/bls12-377/fr/kzg/kzg.go
@@ -77,7 +77,7 @@ func NewSRS(size uint64, bAlpha *big.Int) (*SRS, error) {
 	_, _, gen1Aff, gen2Aff := bls12377.Generators()
 	srs.G1[0] = gen1Aff
 	srs.G2[0] = gen2Aff
-	srs.G2[1].ScalarMultiplication(&gen2Aff, bAlpha)
+	srs.G2[1].ScalarMul(&gen2Aff, bAlpha)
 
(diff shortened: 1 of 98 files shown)


yelhousni:
On gnark side https://github.com/ConsenSys/gnark/pull/342

gbotrel:
Q: why the refactor `ScalarMultiplication` to `ScalarMul` ?

ThomasPiellard (approved):

gbotrel:
Mmhh accidentally merged too fast I think, not that the name change is big, but it impacts user of the lib for no big reasons.

PR #222, opened 2022-07-26
Throw an error when inverse by 0 in `BatchCompressTorus()`

yelhousni:


gbotrel:
So, do we explicitly say that `batchInverse` returns 0 for inv(0)?

yelhousni:
if this gets merged we can keep `inv(0)=0`

gbotrel:
OK but we will need to document BatchInvert independently; since it's a public function it can be used in other places.

PR #225, opened 2022-07-27
batch point conversion: explicitly overwrite (0,0) when point at inf + test

yelhousni:


gbotrel:
maybe add explicit line in function documentation to define behaviour for zero inputs ?

PR #245, opened 2022-10-04
[perf] save 4 muls in ext-Jac add

yelhousni:
The 4 muls needed to check operands equality in ext-Jac add function should be re-used after that.


ivokub (approved):
LGTM

PR #249, opened 2022-10-16
MultiExp in batch affine (draft)

0x0ece:
Hi, we implemented multiexp in affine coordinates using batch inversion and found it 10-20% faster for large N, e.g. N >= 2^22. See Section 5.3: [ia.cr/2022/1396](https://ia.cr/2022/1396).

This PR is not meant to be merged, I just want to present the result and discuss what's the best way to contribute, if you find it interesting. To repro:
```
cd ecc/bls12-377
go test -bench=BenchmarkMultiExpG1ReferenceBatchAffine
```

Here is a bit more context.

Our work started on hardware/FPGA to accelerate large MSM on BLS12-377, but the technique that we developed can be applied to software to implement batch affine on any curve.

The core idea is to reorder point additions into buckets to maximize efficiency. In the paper we call it scheduling.

In rust/python the scheduler would be an iterator that returns points in a different order.
Here's an example of how we propose to impl it in go. This is equivalent to [your loop written as an iterator](https://github.com/0x0ece/gnark-crypto/blob/batch-affine/ecc/bls12-377/multiexp_affine.go#L615-L623), and this is the corresponding ["naive" scheduler](https://github.com/0x0ece/gnark-crypto/blob/batch-affine/ecc/bls12-377/multiexp_affine.go#L559-L592) that returns points in order (this is all throw-away code, just to explain what we did).

The critical point is to impl a scheduler that "doesn't take too much time", yet maximizes the size of the batch used for inversions. In the paper we discuss some of the challenges we encountered, and it's important to note that this approach works well for uniformly random scalars, but performance degrade in edge cases, for example the worst case when the scalars are all the same. Hence the need to discuss, before making a real PR. :)

This is the [actual loop](https://github.com/0x0ece/gnark-crypto/blob/batch-affine/ecc/bls12-377/multiexp_affine.go#L526-L535) that we propose and here's the corresponding [scheduler](https://github.com/0x0ece/gnark-crypto/blob/batch-affine/ecc/bls12-377/multiexp_affine.go#L302). Internally the scheduler has a queue, for simplicity in this code the queue can grow indefinitely, but in a real PR we'd put a hard bound.

Based on our experiments, here's what we'd propose:
1. Implement batch affine only for c=16 and c=20.
2. For large enough N, `msmInnerG1Jac` can use batch affine instead of extended Jacobian. There's a case to be made to keep the code simpler and always use batch affine for c=16, 20, as this pretty much implies large enough N (maybe the improvement is not 10%, but it should still be a few % points).
3. For the batch size, we experimented a bit but at the end landed on [2 fixed values](https://github.com/0x0ece/gnark-crypto/blob/batch-affine/ecc/bls12-377/multiexp_affine.go#L311-L314) T=100 for c=20, T=40 for c=16. It's prob worth testing on the platforms that most interest you.

I know this is a lot :) If you think this is interesting I'm happy to spend some time working on it, and also happy if you prefer to take the lead and structure the code however you think it's best.


gbotrel:
Hey -- looks great, thanks for the PR! @yelhousni and I will probably get back to you in the next week or so, we're unavailable the next few days 👍

0x0ece:
@yelhousni all makes sense. I guess what I can do is to add 2 functions for c=16, 20 with batch affine to the template, so they become available to all curves, G1 and G2. Then you guys can decide if and where to apply them, taking into account all considerations (perf, audit, etc.) - sounds good? BTW, happy to hop on a call if you'd like, maybe faster to sync.

P.S: nice trick the butterfly, love it! You can prob also skip all reductions mod q since you have that multiplier that accepts inputs in [0,2q). Not sure if that has any meaningful impact though (fwiw we used it to save some wires on the FPGA :))

yelhousni:
Yes that would be great if you can do so @0x0ece! Happy to hop on a call too, @gbotrel will be back tomorrow so maybe we can aim for Friday? we can organise this via gnark@consensys.net or [Telegram](https://telegram.me/ElMarroqui).

gbotrel:
@0x0ece couple of updates on our side -- 

* identified an issue in this PR where the scheduler could leave some items in the queue after processing the points
* another one was missing edge case check (if a input point is infinity)
* rewrote a similar (probably temporary version) scheduler which ensures things like buckets are allocated on the stack -- for x86/arm64, results are quite promising -- for G2 MSM the perf gain can be ~30% compared to gnark current version (2**14 -> 2**18). 
* one key observation is with the choice of c (ie the window size in the bucket method); if c doesn't divide the length of the scalar, the last chunk of the scalar will be "smaller" ; that is, we will have much less buckets, and processing this last chunk with the batch-affine version doesn't make sense (due to small batch needed / high number of collisions); what we did is when this is the case, we process the last chunk with buckets in extended jac coordinates. 


All in all, we probably need some time to clean up / find good heuristic for choice of batch size, max batch size, queue strategy, memory accesses, ... for varying sizes of MSM in various settings.

0x0ece:
Thank you for the updates @gbotrel. All looks good to me:
1. Very possible, thanks for spotting and fixing it
2. Yes
3. Awesome!
4. Oh yeah, I forgot to mentioned that during the call, I did the same in my tests but then didn't want to complicate the code. For example [here](https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/bls12-377/multiexp.go#L1103) `c` could be `lastC`. Also, for BLS12-377 you could actually use 253 instead of 256 (therefore tweaking also `msmC16`) and similarly for all curves.

Totally understand this needs some time. I'm glad it's working and you're finding improvements, very exciting!

PR #253, opened 2022-10-28
[perf] `msmReduceChunk` in Jac instead of JacExt

yelhousni:
Currently we implement `msmReduceChunk` in extended Jacobian coordinates. It costs `(s-1)c` doublings, `(s-1)` additions and `1` conversion from JacExt to Jac (`2`sq + `2`mul), where `s` is the chunk size and `c` the window size. For large MSM instances, it is better to do it in Jacobians or at least the doublings in Jacobians, as they cost `3` mul less. The addition, however, costs `2` mul more. This PR trades off `10(s-1)c+14(s-1)+4` muls for `7(s-1)c+16(s-1)+4s` muls.


yelhousni:
Closing this because the benchmarks are not as good on `x86` as on `M1`. e.g. BN254 G1-MSM:
```
M1:
benchmark                          old ns/op      new ns/op      delta
BenchmarkMultiExpG1Reference-8     1336971708     1235571875     -7.58%
```
and
```
AWS z1d.large:
benchmark                          old ns/op      new ns/op      delta
BenchmarkMultiExpG1Reference-2     4399459569     4361069253     -0.87%
```

PR #271, opened 2022-11-18
feat: field.Hash

Tabaie:
For FS commitments it is needed to hash to fr, so I'm moving the `hashToFp` code from the HashToCurve modules into every Fp package.
One concern is that the hash function uses some `ExpandMsgXmd` function, elsewhere in the gnark-crypto codebase. So as it stands field packages can no longer by copy-pasted into other projects without importing gnark-crypto.


gbotrel:
looks good to me 👍 
Small change: package `hash_utils` should be `hashutils` (using Google style guide as reference, but I know some linters complain all the same--> https://google.github.io/styleguide/go/decisions#package-names ) 

We can keep this package in `internal/` for now, but in future it may be merged with other common pieces between field element pacakges 👍

gbotrel (approved):

PR #281, opened 2022-12-13
[clean] endomorphism + code generation 

### LINKED ISSUE #60: Templates should handle correctly the cases where `GLV` and `CofactorCleaning` are set to `false`
e.g. In `./internal/generator/config/bls12-381.go`, if we set `GLV: false` and `CofactorCleaning: false`, the template `internal/generator/ecc/template/point.go.tmpl` handles incorrectly this case.

yelhousni:
fixed #60 
When GLV is set to false (endomorphism not available) in the config file, the templates still generate code for some computations (`ClearCofactorl()`, `IsInSubgroupl()`, `ScalarMul()`) that use the endomorphism `phi()`. This PR fixes the problem.

When the GLV is set to false, the templates:
- use `mulWindowed()` instead of `mulGLV()`
- multiply by `r` for `IsInSubGroup()`
- multiply by the cofactors for `ClearCofactor()` (for BW6 curves)

### DIFF
--- a/internal/generator/ecc/template/point.go.tmpl
+++ b/internal/generator/ecc/template/point.go.tmpl
@@ -9,7 +9,9 @@ import (
 	"math/big"
 	"runtime"
 
+	{{- if .GLV}}
 	"github.com/consensys/gnark-crypto/ecc"
+	{{- end}}
 	"github.com/consensys/gnark-crypto/internal/parallel"
 	"github.com/consensys/gnark-crypto/ecc/{{.Name}}/fr"
 	{{- if or (eq .CoordType "fptower.E2") (eq .CoordType "fptower.E4") }}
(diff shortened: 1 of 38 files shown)


internal/generator/ecc/template/point.go.tmpl
```diff
@@ -796,9 +840,15 @@ func (p *{{$TJacobian}}) ClearCofactor(a *{{$TJacobian}}) *{{$TJacobian}} {
 	L1.AddAssign(&tmp)
 
 	p.phi(&L1).AddAssign(&L0)
+{{ else}}
+    var c1 big.Int
```
gbotrel, line 844: why do we want to allow for a non-GLV path for these curves?

yelhousni: 1- there is a boolean flag that results in errors in the code when set to false. I believe that the first intention was to find a way around the GLV patent (now expired).
2- this should be cleaner way to have other curves without GLV (e.g. STARK curve, NIST P-256..)

gbotrel: ok -- I'ld have taken a shortcut and just panicked for curve we know we have GLV, instead of writing the non-glv path but.. sure :)

PR #282, opened 2022-12-13
Feat/iop arguments

ThomasPiellard:
This PR incorporates the routines which are commonly used in the backend of a 'plonk-like iop', namely:

* Computing the quotient by X^n-1 of a relation of the form h(f_1, .., f_n)=0 on the roots of unity, where h is a multivariate polynomial of n variables and f_i are univariate polynomials of degree n
* Computing the interpolation of accumulating ratios of vectors, interpreted as polynomial in Lagrange basis. There is the version where the ratio Z statisfies Z(w x) =Z(X)*(X-a_1[i])..(X-a_n[i])/(X-b_1[i])..(X-b_n[i]) and Z[0]=1 and the version where the specific permutation is provided (cf paragraph 5 of [plonk](https://eprint.iacr.org/2019/953)).

The functions previously described are already hardcoded in gnark in the plonk backend, but hardcoded and not really reusable. The goal here is to make those functions as generic as possible while keeping the efficient we get when they are hardcoded.
To this end, a new polynomial package has been created where it describes a polynomial as a slice of []fr.Element as we implicitly do, but the structure describing the polynomial also contains:
* its basis (Lagrange, Lagrange shifted, Canonical)
* its status (Locked or Unlocked), useful for not reallocating a slice when it's not needed
* its layout (BitReverse or Regular). When in bitreverse, we do not call fft.BitReverse on it but rather we browse the polynomial in a permuted fashion. This way we avoid as much as possible the costly operation of bit reversing the polynomial.

In view of developing custom gates, a simple structure has also been built to support multivariate polynomials. The structure is just a list of monomials which are composed of a coeff + a list of exponents.


ecc/bls12-377/fr/iop/polynomial.go
```diff
@@ -0,0 +1,165 @@
...
+	"math/big"
+
+	"github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
+)
+
+//-----------------------------------------------------
+// univariate polynomials
+
+// Enum to tell in which basis a polynomial is represented.
+type Basis int64
```
gbotrel, line 30: This enums may find a better home in a dedicated `polynomial` package? 
If we consider a package importing `fr/iop`, they will read as `iop.Canonical` `iop.Regular` `iop.BitReverse` `iop.Locked`, ... which is not super clear.

From what I've seen in the PR so far, I'ld suggest (but maybe I'm wrong) to do a more "private enum" like so:
```
const (  // iota is reset to 0
        canonical = iota  // canonical == 0
        lagrange = iota  // lagrange == 1
        lagrangeCoset = iota  // lagrangeCoset == 2
)
```
And have methods defined on the `Polynomial` (if multiple `Polynomial` needs it make it an interface to implement) like `IsCanonical() bool` etc. Will be inlined by the compiler, and it makes future changes easier to refactor without user impact.

ecc/bls12-377/fr/iop/polynomial.go
```diff
@@ -0,0 +1,165 @@
...
+
+// monomial represents a monomial encoded as
+// coeff*X₁^{i₁}*..*X_n^{i_n} if exponents = [i₁,..iₙ]
+type monomial struct {
+	coeff     fr.Element
+	exponents []int
+}
+
+// it is supposed that the number of variables matches
+func (m monomial) evaluate(x []fr.Element) fr.Element {
```
gbotrel, line 86: why is this not pointer receiver? `m *monomial` would avoid the copy of `m.coeff` at each call.

ecc/bls12-377/fr/iop/polynomial.go
```diff
@@ -0,0 +1,165 @@
...
+// then a new vector is allocated.
+type Status int64
+
+const (
+	Locked Status = iota
+	Unlocked
+)
+
+// Form describes the form of a polynomial.
+type Form struct {
```
gbotrel, line 59: would add methods here `form.SetBasis()` `form.Lock()` .. `form.Unlock()` . Then in case debugging is needed, it's easy to add some (temporary or build tag enabled) consistency checks in these APIs, rather than changing the state through `polynomial.info.status = lock`

ecc/bls12-377/fr/iop/polynomial.go
```diff
@@ -0,0 +1,165 @@
...
+	Basis  Basis
+	Layout Layout
+	Status Status
+}
+
+// Polynomial represents a polynomial, the vector of coefficients
+// along with the basis and the layout.
+type Polynomial struct {
```
gbotrel, line 67: maybe embed the `Form` 
```
type Polynomial struct {
    Form
    Coefficients ...
}
```

Then you have nice syntax; 
```
var p Polynomial
p.Lock()
p.SetBasis(...)
p.IsBitReversed(...)
```

ecc/bls12-377/fr/iop/quotient.go
```diff
@@ -0,0 +1,176 @@
...
+func ComputeQuotient(entries []*Polynomial, h MultivariatePolynomial, expectedForm Form, domains [2]*fft.Domain) (Polynomial, error) {
+
+	var quotientLagrangeCosetBitReverse Polynomial
+
+	// check that the sizes are consistant
+	err := checkSize(entries)
+	if err != nil {
+		return quotientLagrangeCosetBitReverse, nil
+	}
+
```
gbotrel, line 38: add a check that domains[...] are correct; for example that domains[n].Cardinality != 0 ?

ecc/bls12-377/fr/iop/ratios.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	// at this stage the result is in Lagrange form, Regular layout
+	putInExpectedFormFromLagrangeRegular(&res, domain, expectedForm)
+
+	return res, nil
+
+}
+
+func putInExpectedFormFromLagrangeRegular(p *Polynomial, domain *fft.Domain, expectedForm Form) {
```
gbotrel, line 241: should you have a single method to convert to expected form ? 
Like 
```
if self.IsBitReversed() && !expected.BitReversed() {
    self.BitReverse(...)
}
// etc.
```
?

ecc/bls12-377/fr/iop/utils.go
```diff
@@ -0,0 +1,378 @@
...
+func printPolynomials(p []*Polynomial) {
+	fmt.Printf("[\n")
+	for i := 0; i < len(p); i++ {
+		printVector(p[i].Coefficients)
+		fmt.Printf(",\n")
+	}
+	fmt.Printf("]\n")
+}
+
+func printLayout(f Form) {
```
gbotrel, line 43: add a `String()` method to the enums and a `String()` method to the `Form` type

ecc/bls12-377/fr/iop/utils.go
```diff
@@ -0,0 +1,378 @@
...
+	} else {
+		fmt.Printf("UNLOCKED")
+	}
+	fmt.Println("")
+}
+
+type modifier func(p *Polynomial, d *fft.Domain) *Polynomial
+
+// return a copy of p
+func copyPoly(p Polynomial) Polynomial {
```
gbotrel, line 72: add `Polynomial.Clone() Polynomial` method?

ecc/bls12-377/fr/iop/utils.go
```diff
@@ -0,0 +1,378 @@
...
+
+func exp5(x fr.Element) fr.Element {
+	var res fr.Element
+	res.Square(&x).Square(&res).Mul(&res, &x)
+	return res
+}
+
+// doesn't return any errors, it is a private method, that
+// is assumed to be called with correct arguments.
+func smallExp(x fr.Element, n int) fr.Element {
```
gbotrel, line 358: unused? 
if used, I think it's less verbose to init 5 big int to avoid many allocations, do a switch on n and call fr.Element.Exp(.., bigINt), perf should be OK.

ecc/bls12-377/fr/iop/utils.go
```diff
@@ -0,0 +1,378 @@
...
+	r.Info = p.Info
+	return r
+}
+
+// return an ID corresponding to the polynomial extra data
+func getShapeID(p Polynomial) int {
+	return int(p.Info.Basis)*4 + int(p.Info.Layout)*2 + int(p.Info.Status)
+}
+
+//----------------------------------------------------
```
gbotrel, line 86: mmhhh I don't like that part it doesn't seem really robust to changes; 
I think the combinatorial is not too crazy and this can be factored in a single method; 
```
isLocked := p.IsLocked()
if isLocked {
    p = p.Clone()
}
if targetBasis != p.Basis {
    // ...
}
```

at the minimum, would be nice to have some compiler-enforced consistency checks / safety.

ecc/bls12-377/fr/iop/utils_test.go
```diff
@@ -0,0 +1,784 @@
...
+	// get a backup
+	backupPolynomials := getCopy(polynomials)
+
+	// create the Lagrange form...
+	lagrangePolynomials := make([]Polynomial, 12)
+	for i := 0; i < 12; i++ {
+		lagrangePolynomials[i] = *toLagrange(&polynomials[i], domain)
+	}
+
+	// compare the results that should be in regular form
```
gbotrel, line 297: I don't know how to review that ^^

gbotrel (changes requested):
looks good, didn't really check the correctness of test, but one thing that bother me is the `Form` metadata associated with a `Polynomial` and the way the conversions are implemented / tested, I think we can do better to ensure consistency / safety of the APIs.

ecc/bls12-377/fr/iop/polynomial.go
ThomasPiellard: I agree on the polynomial package, it's just that currently on Arya's branch for gkr an existing polynomial package is already used, but the plan is to eventually replace it. I'll talk with Arya about that

ecc/bls12-377/fr/iop/utils.go
ThomasPiellard: Yeah those were for debugging that I forgot to delete

ecc/bls12-377/fr/iop/utils_test.go
ThomasPiellard: yeah no worries those are just utility functions that I checked with Sage

PR #284, opened 2022-12-14
Replace partitionScalarOld by partitionScalar in batchScalarMul

gbotrel:
- checkpoint
- feat: replace partitionScalarOld by partitionScalar


gbotrel:
ha, well, I think this new `partitionScalars` would work as is; the returned values is a list of `[]uint16 digits`, so normally we would just have a last window that's larger than c. The only issue is that... if we have c = 16 and lastC = 17, the digits don't fit on `uint16` 🙃

Issue #276, opened 2022-12-05
What are the usecases for ToBigInt method?


kevaundray:
**Motivation**

When I think about the public methods on a field struct, I usually assume that they do not leak any internal implementation detail. In this case, the fact that Montgomery multiplication/representation is being used instead of Barret Multiplication for example, seems to be being leaked by ToBigInt which returns the element in montgomery form as a golang big integer.

**Discussion**

It then prompted me to ask, if there are any usecases for having this exposed that I'm not aware of?

The main downside I can see is that, as an external user, if you do not know what montgomery form is and or you are not reading the documentation, you may call `ToBigInt` thinking that it's behaviour is akin to `ToBigIntRegular`.

**Limitations Imposed**

If the fact that montgomery is being used is not exposed through the public API, then this means that the following invariants will hold:

- All big integers created from a field element, will be in "regular" form
- Creating a field element can only be done via either copying another field element or inputting a byte array/integer which is in "regular" form.

- `ToBigInt` would then be `toBigInt`
- `FromMont` which converts a field element from montgomery form to regular form, would most likely be removed. (Meaning there would be no public methods which violate the rule that all field elements are internally in montgomery form)

**Main arguments against**

The whole post assumes that the user is not experienced and likely to shoot themselves in the foot, this may not be the intended target audience and the API may be good just the way it is :)

kevaundray:
I think this line of reasoning would also apply to `FromMont` being in the public API. 

It's also likely that this is the only interface that works due to language constraints :)

gbotrel:
At first sight, I think yes, deprecating `ToBigInt` (to avoid behavior change with lib update) may make sense. Probably here for historical reasons.

`FromMont` on the elements however is needed; we may work at the bit level with field elements in various places, and need the element to be represented in canonical form to do so.

kevaundray:
> 

I would naively suggest a `toBytes` or `toBits` method here, but I guess you are also worried about perf?

gbotrel:
We might work around it -- though indeed, doing it in place helps with perf. 
I suspect going the `ToBits()` will add more code than it will remove accros the code base.

kevaundray:
> We might work around it -- though indeed, doing it in place helps with perf.
> I suspect going the `ToBits()` will add more code than it will remove accros the code base.

Makes sense, I will lean on your knowledge of the codebase here :)

gbotrel:
after a second read on this, I'm more convinced that this is something we should do :) 
[...]

PR #288, opened 2022-12-16
Removes `FromMont` and `ToMont` from `field.Element` api

gbotrel:
See #276


ecc/bls12-381/hash_vectors_test.go
```diff
@@ -9,6 +9,7 @@ import (
 func TestG1IsogenyVectors(t *testing.T) {
 	t.Parallel()
 
+	// TODO @gbotrel fix me test vectors shouldn't set words directly
```
Tabaie, line 12: Right, if these come from the standard docs they should be available in decimal or hex form.

Tabaie (approved):

samlaf:
Trying to upgrade our version of gnark-crypto from 0.8 to 0.9.
We were using
```
config := ecc.MultiExpConfig{ScalarsMont: true}
```
whereas now MultiExpConfig does not have a ScalarsMont argument.

I see that the default was set to false in 0.8. Can I just remove that field and everything will work fine? Or will this lead to breaking changes?

Disclaimer: I don't know what montgomery form is.

gbotrel:
That was actually an overlook on my sight, pointed by other users in the past;

there is actually a subtle breaking change here.. 

the default is now effectively `ecc.MultiExpConfig{ScalarsMont: true}` so if you used that everywhere, you should be fine.

But callers without an explicit value would have had the default at: `ecc.MultiExpConfig{ScalarsMont: false}` and  the upgrade would be a silent change.. --> TLDR; all the APIs are now consistant and don't expose the underlying form / encoding of the field elements.

samlaf:
> That was actually an overlook on my sight, pointed by other users in the past;
> 
> there is actually a subtle breaking change here..
> 
> the default is now effectively `ecc.MultiExpConfig{ScalarsMont: true}` so if you used that everywhere, you should be fine.
> 
> But callers without an explicit value would have had the default at: `ecc.MultiExpConfig{ScalarsMont: false}` and the upgrade would be a silent change.. --> TLDR; all the APIs are now consistant and don't expose the underlying form / encoding of the field elements.

Awesome thanks! Small note: would have been useful if you guys added a `Breaking Changes` section to the release notes.
Otherwise is there a good reference that you would recommend to understand these montgomery forms and how they are used in this library?

Issue #242, opened 2022-09-27
Pointer/non-Pointer refceivers


Tabaie:
Struct methods can take either pointer or value receivers. We have structs with methods of both types, which is now discouraged by Go documentation. We should either fix all such instances or have an argument why not.

kevaundray:
Just out of curiosity, can you explain what you mean when you say, _we have structs with methods of both types_ ?

My incorrect interpretation of this sentence is:

```go
type Foo uint64

func (f Foo) hello()  {}
func (f *Foo) hello() {}
```

Incorrect because the above would not compile.

Tabaie:
I guess it is better to say of _each_ type:

```go
type Foo uint64

func (f Foo) hello() {}
func (f *Foo) goodbye() {}
```

kevaundray:
> I guess it is better to say of _each_ type:
> 
> ```go
> type Foo uint64
> 
> func (f* Foo) hello()  {}
> func (f Foo) goodbye() {}
> ```

Ah thanks for clarifying. To add more context to this issue, I found [this article](https://nathanleclaire.com/blog/2014/08/09/dont-get-bitten-by-pointer-vs-non-pointer-method-receivers-in-golang/) which also corroborates your position.

Tabaie:
Makes sense, thank you!

ivokub:
Seems stale?

PR #290, opened 2022-12-19
refactor: mark ToBigIntRegular as deprecated, introduce BigInt method

gbotrel:


field/generator/internal/templates/element/conv.go
```diff
@@ -83,7 +83,15 @@ func (z *{{.ElementName}}) Text(base int) string {
 	{{- end}}
 }
 
+// BigInt sets and return z as a *big.Int
+func (z {{.ElementName}}) BigInt(res *big.Int) *big.Int {
```
Tabaie, line 87: Using mixed receivers (pointer for some functions and regular for others) is discouraged. Better to explicitly copy.
https://github.com/ConsenSys/gnark-crypto/issues/242

Tabaie (changes requested):

PR #292, opened 2022-12-23
Avoid unnecessary big.Int assignment

jsign:
This PR removes an unnecessary assignment to an output variable.


internal/field/internal/templates/element/conv.go
```diff
@@ -159,7 +159,6 @@ func (z *{{.ElementName}}) SetBigInt(v *big.Int) *{{.ElementName}} {
 	vv := bigIntPool.Get().(*big.Int)
 
 	// copy input + modular reduction
-	vv.Set(v)
```
jsign, line 162: `vv` will be used as an output variable in the operation next line, so it shouldn't make

gbotrel:
cherry picked in #296

PR #299, opened 2023-01-09
Add STARK curve

yelhousni:
Supports only Fp, Fr and G1 arithmetics.
TODO:
- [ ] Optimise field arithmetic given the special form of the moduli
- [ ] Use addchain for Sqrt and Legendre (bug with addchain --> code generation works only when `NewField()` set with false flag)
----
- [ ] Code generation for G1 when `a!=0` (but not sure we'll add another curve with `a!=0`)
- [ ] MultiExp? (but not needed for now, maybe for Pedersen hash or batch ECDSA?)


gbotrel (approved):
lgtm 👍 
on the TODOs:
1. addchain works OK with fp. Not fr, I think it's a small bug in `addchain` but need to write a minimal reproductible test case... (it finds the addition chain for the legendre exponent, but the code generation template fails)
2. field arithmetic: since q0 == 1, q1 == 0, q2 == 0, many functions can be optimized, but the arithmetic templates are dirty and need some refactoring to allow that cleanly. Probably couple of days work and a good double digit perf boost on the fp.Mul (and consequently all the ecc ops cc @joshklop @omerfirmak )

omerfirmak:
> field arithmetic: since q0 == 1, q1 == 0, q2 == 0, many functions can be optimized, but the arithmetic templates are dirty and need some refactoring to allow that cleanly. Probably couple of days work and a good double digit perf boost on the fp.Mul (and consequently all the ecc ops cc @joshklop @omerfirmak )

Pedersen is still the main bottleneck in our application, so that would be much appreciated if you can find the time. Thank you for all the work so far!

yelhousni:
Let's merge this and track the field arithmetic optimization in #302

Issue #304, opened 2023-01-12
Field-agnostic Fiat-Shamir Challenge Names


Tabaie:
1. Currently, in order to align hash inputs correctly, the user of a Fiat-Shamir transcript has to pad the challenge names so that they are as long as a hashing block. It is desirable for that to be done automatically, 

2. It is better to pre-pad rather than post-pad with zeros to avoid the resulting block being "larger" than the field modulus. This will make `gnark/std/fiat-shamir.TestFiatShamir` pass (it currently fails.)

[304](https://github.com/ConsenSys/gnark-crypto/tree/304-field-agnostic-fiat-shamir-challenge-names) addresses these issues, but introduces another: `gnark/std/commitments/fri.TestFriVerification` fails because a challenge name is too large. To solve this we must somehow pass a decompose function to the transcript object, such that for arithmetic hashes it operates as introduced in https://github.com/ConsenSys/gnark-crypto/pull/265 and for traditional hashes it simply breaks them up into blocks. How to do this cleanly such that `Transcript` remains hash-agnostic and `NewTranscript` doesn't get more complicated? I suggest instead of using the standard go `Hash` interface we use a wrapper that also has a `Decompose` function.

There are other failing `gnark` tests (with `gnark-crypto@develop`) that this issue doesn't address:
* `gnark/std/accumulator/merkle.TestVerify`
* `gnark/std/commitments/fri.TestFriVerification`
* `gnark/std/signature/eddsa.TestEddsa`

Tabaie:
@gbotrel @ThomasPiellard

gbotrel:
So, to sum up state of things after your PR, we would have an arithmetic hash like MiMC that is supposed to take as input field elements, that would implement the following 2 interfaces;

```go
type ArithmeticHash interface {
 	WriteString(rawBytes []byte) error
 }
```
 and from go std:
 ```go
 type Hash interface {
	// Write (via the embedded io.Writer interface) adds more data to the running hash.
	// It never returns an error.
	[io](https://pkg.go.dev/io).[Writer](https://pkg.go.dev/io#Writer)
	// ...
}
```

I think #265 introduced a bad decision; make `Write` return an error if the `[]byte` input is not mod reduced. 

Since it's cumbersome to add `fr.Element` in signatures here, what if we only keep `Write([]byte)`, have a fast path (if input is mod reduced, no decomposition involved).

The piece we need to document well is; as a user, if I do:
```go
h.Write(0xAA)
[...]

PR #308, opened 2023-01-16
304 field agnostic fiat shamir challenge names

Tabaie:
This PR proposes a solution to #304 by introducing an interface called `ArithmeticHash` with a `WriteString` function that decomposes its message before writing it. Used for long fiatshamir challenge names and signatures.

THE ABOVE IS OUTDATED: We decided on a more lightweight solution: the idea of keeping the hash interface but returning error on non-field-element input is held on to, but the decompose method is replaced with hashes to field.


ecc/bls12-377/fr/mimc/mimc.go
```diff
@@ -124,14 +123,14 @@ func (d *digest) checksum() fr.Element {
 	// Write guarantees len(data) % BlockSize == 0
 
 	// TODO @ThomasPiellard shouldn't Sum() returns an error if there is no data?
-	if len(d.data) == 0 {
+	// TODO: @Tabaie, @Thomas Piellard Now sure what to make of this
```
gbotrel, line 126: 2 solutions I guess; 
1. since `checksum` is called in `Sum` and `Sum` can return an error, we could do that if there is no data to hash.
2. refer to other Hash impl in Go std and see how Sum behaves with empty buffer.

ecc/bls12-377/fr/mimc/mimc.go
```diff
@@ -98,23 +98,22 @@ func (d *digest) BlockSize() int {
 // If len(p) is not a multiple of BlockSize and any of the []byte in p represent an integer
 // larger than fr.Modulus, this function returns an error.
 //
```
gbotrel, line 100: add in godoc a warning to say that this method may error (whereas Hash interface{} say it may not)

hash/to_field.go
```diff
@@ -0,0 +1,5 @@
+package hash
+
+type ToField interface {
+	WriteString(rawBytes []byte)
```
gbotrel, line 4: I think this file is not needed; just do a private interface{} in `transcript.go`

gbotrel (approved):

ecc/bls12-377/fr/mimc/mimc.go
Tabaie: Took a look at SHA256, turns out for conventional hashes hashing nothing makes sense, and provides a valid input. I think our choice is between returning 0 or throwing an error.

PR #310, opened 2023-01-18
feat: Add ECDSA

yelhousni:
Adds ECDSA scheme using gnark-crypto primitives on all curves. This follows "crypto/ecdsa", FIPS 186-4 and SEC1 v2.

todos:
- [ ] use `fr.Inverse()` instead of `bigInt.ModInverse()` (experimented speedup is just ~1-2% so need to dig more why)
- [x] Implement Strauss-Shamir trick `[a]P + [b]Q` for fast verification (might also look into JSF encoding of scalars if perf is critical)


yelhousni:
- [x] Strauss-Shamir trick (without GLV)

ivokub (changes requested):
Looks good. I think you should copy input scalars to JointScalarMultiplication. In usual ScalarMul, you get the scalars from GLV decomposition and they are safe to modify locally. For ECDSA it is safe as you create big.Int inputs freshly, but if anyone would use JointScalarMul separately, then it would mutate its inputs.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -0,0 +1,219 @@
...
+
+// PrivateKey represents an ECDSA private key
+type PrivateKey struct {
+	PublicKey
+	Secret big.Int
+}
+
+// Signature represents an ECDSA signature
+type Signature struct {
+	R, S big.Int
```
ivokub, line 25: Usually the convention is to use *big.Int. There are slices in the struct anyway, so is always heap-allocated.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -0,0 +1,219 @@
...
+
+// PublicKey represents an ECDSA public key
+type PublicKey struct {
+	Q {{ .CurvePackage }}.G1Affine
+}
+
+// PrivateKey represents an ECDSA private key
+type PrivateKey struct {
+	PublicKey
+	Secret big.Int
```
ivokub, line 20: Use *big.Int (by convention)

ivokub (commented):
To record conversation - maybe it is better to use `fr.Element` in `Signature` and `PrivateKey` types.

Also, it seems `params` type isn't strictly necessary - we have all the information in the curve template anyway. And add add `signature/ecdsa` package which calls the actual implementation depending on the curve.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -0,0 +1,287 @@
...
+			if r.Sign() != 0 {
+				break
+			}
+		}
+		s.Mul(r, scalar)
+
+		// compute the hash of the message as an integer
+		dataToHash := make([]byte, len(message))
+		copy(dataToHash[:], message[:])
+		hFunc.Reset()
```
ivokub, line 217: signature.Signer interface says that if hFunc is nil then implementation may assume pre-hashed message. I think it would make sense to also consider this case.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -0,0 +1,287 @@
...
+ 	}
+
+	r, s := new(big.Int), new(big.Int)
+	r.SetBytes(sig.R[:sizeFr])
+	s.SetBytes(sig.S[:sizeFr])
+
+	sInv := new(big.Int).ModInverse(s, order)
+
+	// compute the hash of the message as an integer
+	dataToHash := make([]byte, len(message))
```
ivokub, line 260: Similarly, should consider when hFunc is nil.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -0,0 +1,287 @@
...
+//
+// SEC 1, Version 2.0, Section 4.1.3
+func (privKey *PrivateKey) Sign(message []byte, hFunc hash.Hash) ([]byte, error) {
+	scalar, r, s, kInv := new(big.Int), new(big.Int), new(big.Int), new(big.Int)
+	scalar.SetBytes(privKey.scalar[:sizeFr])
+	for {
+		for {
+			csprng, err := nonce(privKey, message)
+			if err != nil {
+				return nil, err
```
ivokub, line 195: Usually, when a method has different paths for returning errors, then I wrap the error with `fmt.Errorf("error source: %w", err)`, which wraps the error with something descriptive. Makes a little easier to debug, but is not that important here because we use stack traces anyway.

ivokub (approved):
LGTM

internal/generator/ecdsa/template/ecdsa.go.tmpl
yelhousni: This is the case of eddsa too. Should we merge this and do a PR for both ecdsa and eddsa?

ivokub: Yep, makes sense.

yelhousni: done in `signature/pre-hashed` branch. Will push the small PR when this one is merged.

PR #311, opened 2023-01-24
feat: introduce field.Vector

gbotrel:
each `field`  package now exposes a `Vector` (`[]Element`) object, which implements the following interfaces:
```go
//   - Stringer
//   - io.WriterTo
//   - io.ReaderFrom
//   - encoding.BinaryMarshaler
//   - encoding.BinaryUnmarshaler
//   - sort.Interface
```

It replaces the `plookup.Table` object in `gnark-crypto`, should probably use in `ecc/Encoder` functions, and will be the underlying representation of a zkSNARK witness in `gnark`.  

- refactor: element imports field/pool and field/hash to avoid cycles
- feat: adds Vector in field package
- refactor: plookup.Table -> fr.Vector


ecc/bls12-377/fp/element_test.go
```diff
@@ -2313,7 +2325,7 @@ func (z *Element) matchVeryBigInt(aHi uint64, aInt *big.Int) error {
 
 	slice := append(z[:], aHi)
 
-	return field.BigIntMatchUint64Slice(&aIntMod, slice)
+	return bigIntMatchUint64Slice(&aIntMod, slice)
```
Tabaie, line 2328: Maybe we can start getting rid of some of the tests for subcomponents of the Pornin inverse algo not that it's been a while and it seems to work.

Tabaie (approved):

ecc/bls12-377/fp/element_test.go
gbotrel: yep, even more so that... the pornin Inverse has this "fallback" mechanism to call the inverseExp.

PR #312, opened 2023-01-25
pre-hashed message in ECDSA and EdDSA

yelhousni:
Following the signature interface, for both ECDSA and EdDSA:
```
// If hFunc is not provided, implementation may consider the message
// to be pre-hashed, else, will use hFunc to hash the message.
```

### DIFF
--- a/internal/generator/ecdsa/template/ecdsa.go.tmpl
+++ b/internal/generator/ecdsa/template/ecdsa.go.tmpl
@@ -211,16 +211,21 @@ func (privKey *PrivateKey) Sign(message []byte, hFunc hash.Hash) ([]byte, error)
 		}
 		s.Mul(r, scalar)
 
-		// compute the hash of the message as an integer
-		dataToHash := make([]byte, len(message))
-		copy(dataToHash[:], message[:])
-		hFunc.Reset()
-		_, err := hFunc.Write(dataToHash[:])
-		if err != nil {
-			return nil, err
-		}
-		hramBin := hFunc.Sum(nil)
-		m := HashToInt(hramBin)
+		m := new(big.Int)
+		if hFunc != nil {
+			// compute the hash of the message as an integer
+			dataToHash := make([]byte, len(message))
+			copy(dataToHash[:], message[:])
+			hFunc.Reset()
+			_, err := hFunc.Write(dataToHash[:])
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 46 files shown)


internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -256,16 +261,21 @@ func (publicKey *PublicKey) Verify(sigBin, message []byte, hFunc hash.Hash) (boo
...
-	dataToHash := make([]byte, len(message))
-	copy(dataToHash[:], message[:])
-	hFunc.Reset()
-	_, err := hFunc.Write(dataToHash[:])
-	if err != nil {
-		return false, err
-	}
-	hramBin := hFunc.Sum(nil)
-	m := HashToInt(hramBin)
+	m := new(big.Int)
```
ivokub, line 264: It is sufficient here to do `var m *big.Int` and not allocate. You assign value to m later. This allocation will be wasted.

internal/generator/ecdsa/template/ecdsa.go.tmpl
```diff
@@ -211,16 +211,21 @@ func (privKey *PrivateKey) Sign(message []byte, hFunc hash.Hash) ([]byte, error)
...
-		dataToHash := make([]byte, len(message))
-		copy(dataToHash[:], message[:])
-		hFunc.Reset()
-		_, err := hFunc.Write(dataToHash[:])
-		if err != nil {
-			return nil, err
-		}
-		hramBin := hFunc.Sum(nil)
-		m := HashToInt(hramBin)
+		m := new(big.Int)
```
ivokub, line 214: It is sufficient here to do `var m *big.Int` and not allocate. You assign value to m later. This allocation will be wasted.

ivokub (changes requested):

ivokub (approved):
Looks good and I think is good to merge as-is. Your call if you want to merge currently as-is and have later optional PR if have figured out #322 or wait and incorporate changes.

ivokub:
> > Looks good and I think is good to merge as-is. Your call if you want to merge currently as-is and have later optional PR if have figured out #322 or wait and incorporate changes.
> 
> I think we can merge now as-is since the two other PRs #313 and #314 are based on this branch. Then we can have a PR that refactor all of the 4 schemes (EdDSA, ECDSA, Schnorr, BLS) in a similar manner as per #322.

Makes sense and I agree.

PR #313, opened 2023-01-25
Feat: Add Schnorr signature scheme

yelhousni:


ecc/bls12-377/signature/schnorr/schnorr.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	k, err := randFieldElement(csprng)
+	if err != nil {
+		return nil, err
+	}
+
+	var P bls12377.G1Affine
+	P.ScalarMultiplicationBase(k)
+
+	if hFunc != nil {
```
ivokub, line 210: Hmm, I think this is incorrect. In Schnorr we use the hash for Fiat-Shamir challenge. If `hFunc==nil`, then we do not hash signer randomness into challenge and get a weak instance. Depending on the attack, we either leak secret key or allow forging signatures.

In description, we should rather think as the challenge as `H1(P || H2(m))`, where `hFunc` refers to `H2` not `H1`. `H1` should be fixed per-curve to fill the whole scalar element.

ecc/bls12-377/signature/schnorr/schnorr.go
```diff
@@ -0,0 +1,286 @@
...
+	r, s, e := new(big.Int), new(big.Int), new(big.Int)
+	r.SetBytes(sig.R[:sizeFr])
+	s.SetBytes(sig.S[:sizeFr])
+
+	var P bls12377.G1Jac
+	P.JointScalarMultiplicationBase(&publicKey.A, s, r)
+	var _P bls12377.G1Affine
+	_P.FromJacobian(&P)
+
+	if hFunc != nil {
```
ivokub, line 263: See comment about Fiat-Shamir in Sign.

ivokub (changes requested):
There is a key leakage/forgeability problem imo.

ecc/bls12-377/signature/schnorr/schnorr.go
ivokub: I thought about it. I think the secret key is safe, but still allows forgeability.

yelhousni: I think there is no point into hashing the message (H2) before hashing its concatenation with P (H1), as we consider arbitrary long messages. My initial thought was that if the user uses `nil` as `hFunc` it would mean he did `H(P || m)` prior to signing, and we just convert it to and integer. But I agree that if it opens room to ambiguity we should clear this up. So either:
- We remove the `hFunc` argument from the interface and enforce a hash function inside `Sign()` and `Verify()` for all signature schemes --> but we wouldn't be able to replace with a SNARK-friendly hash in gnark if needed, or
- if `hFunc == nil` then we use an encoded hash inside `Sign()` and `Verify()`, otherwise we use the `hFunc` provided by the caller --> but we would always compute a hash and can't delegate this (e.g. out-circuit in SNARKs), or
- `hFunc` would mean H1 as in your example and we always do H2 --> same we can't delegate H2 but the interface and implementation would be similar across signature schemes.

ivokub: One option is if there is some nice SNARK-friendly instantiation of the Fiat-Shamir challenger for Schnorr signatures. https://eprint.iacr.org/2020/915.pdf says there is (Theorem 2.4 and Construction 2.5), but as I understand, imo requires a (one-time?) trusted setup. But if it is true, then we can fix `H1` and `hFunc` gives `H2`.

yelhousni:
> Does it makes sense? If so, then I would create a new issue to design it further.

Yup that makes sense.

0xy0k:
In this route, I guess, it's also worth to think about standardizing an interface or struct (e.g. SignatureParams) that bundles H1, H2, and any SNARK-friendly hash flags. Could help simplify key serialization too.

gbotrel:
Review pass from 2026-05-07.

Findings:
- Blocking: GitHub currently reports this PR as `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and there is no current CI signal on the PR head.
- Blocking: the existing `CHANGES_REQUESTED` review from 2023 is still unresolved and explicitly calls out a key-leakage/forgeability concern. The later discussion also points toward redesigning the signature parameters as key-level configuration instead of passing hash functions/options at each `Sign`/`Verify` call. I would not treat this as mergeable until that cryptographic/API issue is resolved, the branch is rebased, and CI is restored.

Verification run locally:
- `go test ./signature/...` (top-level signature packages have no tests)
- `go test ./ecc/.../signature/...`

PR #319, opened 2023-01-27
Perf/gkr

Tabaie:


ecc/bls12-377/fr/polynomial/pool.go
```diff
@@ -44,7 +44,7 @@ type inUseData struct {
 
 type Pool struct {
 	//lock     sync.Mutex
-	inUse    map[*fr.Element]inUseData
+	inUse    sync.Map
```
gbotrel, line 47: note in general about the memory pool; impact on benchmarks on my M1 laptop where negligible (vs just calling `make([]fr.Element, ..)`)) . If it has no impact on the `hpc6a` instance, I'ld rather we remove this, the only piece in our codebase with `unsafe` import.

gbotrel (approved):

ecc/bls12-377/fr/polynomial/pool.go
Tabaie: I think its impact will increase the larger the number of instances are. Will check.

PR #320, opened 2023-01-30
fix: number of rounds for mimc

yelhousni:


yelhousni:
For bls12-377, I changed it to:
- permutation exponent `e=17` (instead of -1)
- number of rounds `r=62` (`log(r)/log(17`)

This follows Penumbra reasoning ([doc](https://protocol.penumbra.zone/main/crypto/poseidon/paramgen.html), [audit](https://research.nccgroup.com/wp-content/uploads/2022/09/NCC_Group_PenumbraLabsInc_Report_2022-09-12_v1.1.pdf)) for the Poseidon hash. Out-circuit `x^17` can be done in 4 squaring and 1 multiplication (much faster than an inverse) and in-circuit in 5 R1CS constraints per round (`62*5=310`), which is faster than BN254 (`110*3=330`) and BLS12-381 (`111*3=333`).

Corresponding PR on gnark side: https://github.com/ConsenSys/gnark/pull/453

ThomasPiellard (approved):

PR #323, opened 2023-01-31
321 bn254mimc test vector

Tabaie:
Simple test vector generator for MiMC - BN254.

Possible other features:

- Include result by round (or every 10 rounds so that the test entries are smaller)
- Include test cases with custom string input (conventionally hashed to fr first)


gbotrel:
looks good, may be worth it to have at least one vector that takes multiple elements as input. I believe the intended use case is state manager so I hear "merkle tree path" 👍

PR #324, opened 2023-02-01
feat: pedersen_hash added to stark curve

anshulforyou:
Added the pedersen hash to the stark curve ecc family

### DIFF
--- a/ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
+++ b/ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
@@ -1,0 +1,85 @@
+package pedersenhash
+
+import (
+	"math/big"
+
+	starkcurve "github.com/consensys/gnark-crypto/ecc/stark-curve"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+var (
+	shiftPoint *starkcurve.G1Jac
+	p0         *starkcurve.G1Jac
+	p1         *starkcurve.G1Jac
+	p2         *starkcurve.G1Jac
+	p3         *starkcurve.G1Jac
+)
+
+func init() {
+	// The curve points come from the [reference implementation].
+	//
+	// [reference implementation]: https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/crypto/signature/fast_pedersen_hash.py
+
+	x := new(fp.Element)
+	y := new(fp.Element)
+
+	x.SetString("2089986280348253421170679821480865132823066470938446095505822317253594081284")
+	y.SetString("1713931329540660377023406109199410414810705867260802078187082345529207694986")
+	shiftPoint = new(starkcurve.G1Jac).FromAffine(&starkcurve.G1Affine{X: *x, Y: *y})
+
+	x.SetString("996781205833008774514500082376783249102396023663454813447423147977397232763")
+	y.SetString("1668503676786377725805489344771023921079126552019160156920634619255970485781")
+	p0 = new(starkcurve.G1Jac).FromAffine(&starkcurve.G1Affine{X: *x, Y: *y})
+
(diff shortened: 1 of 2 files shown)


yelhousni:
Thank you @anshulforyou for you contribution! Actually, for other curves, the Pedersen hash we would be interested in would be on the companion twisted Edwards curves (`twistededwards` package). So either we move your Pedersen hash to the root `stark-curve/` in which case the other Pedersen hashes would be in `twistededwards/` (as in eddsa) or we move it to `fp/` in which case the others would be in `fr/` (which is the `fp/` of these companion twisted Edwards). 
Anyway, let me review your PR and get it merged so that you can guys use it and then we'll figure out for the other curves.

ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
```diff
@@ -0,0 +1,85 @@
...
+
+var (
+	shiftPoint *starkcurve.G1Jac
+	p0         *starkcurve.G1Jac
+	p1         *starkcurve.G1Jac
+	p2         *starkcurve.G1Jac
+	p3         *starkcurve.G1Jac
+)
+
+func init() {
```
yelhousni, line 18: Maybe move these to `stark_curve.go` in `init()` there.

ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
```diff
@@ -0,0 +1,85 @@
...
+//
+// [Pedersen hash]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
+// [reference implementation]: https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/crypto/signature/fast_pedersen_hash.py
+func Pedersen(a *fp.Element, b *fp.Element) *fp.Element {
+	result := new(starkcurve.G1Jac).Set(shiftPoint)
+	result.AddAssign(processElement(a, p0, p1))
+	result.AddAssign(processElement(b, p2, p3))
+
+	var affine starkcurve.G1Affine
+	affine.FromJacobian(result)
```
yelhousni, line 68: We just need the `X` coordinate here. So maybe instead of calling `FromJacobian()` you recover the `X` coordinate alone. This would save 2 muls.

ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
```diff
@@ -0,0 +1,85 @@
...
+
+func processElement(a *fp.Element, p1 *starkcurve.G1Jac, p2 *starkcurve.G1Jac) *starkcurve.G1Jac {
+	var bigInt big.Int
+	var aBytes [32]byte
+	a.BigInt(&bigInt).FillBytes(aBytes[:])
+
+	highPart := bigInt.SetUint64(uint64(aBytes[0])) // The top nibble (bits 249-252)
+	lowPart := aBytes[1:]                           // Zero-out the top nibble (bits 249-252)
+
+	m := new(starkcurve.G1Jac).ScalarMultiplication(p2, highPart)
```
yelhousni, line 80: Instead of computing `highPart * p2` and `lowPart * p1` independently and then sum them, you can use `JointScalarMultiplication()` which does the same but faster using Strauss-Shamir trick (see https://github.com/ConsenSys/gnark-crypto/pull/310).

yelhousni (changes requested):
I think you can move the `pedersen_hash/` to the root dir `stark_curve/` for now. The other perf-related changes are optional.

ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
yelhousni: #327

ecc/stark-curve/fr/pedersen_hash/pedersen_hash.go
yelhousni: #327

yelhousni:
continues in #327

PR #327, opened 2023-02-03
Pedersen hash on stark-curve

yelhousni:
#324


yelhousni:
@anshulforyou are the test vectors in the test file generated by the same code here?

yelhousni:
Some optimisations in mind:
- [ ] the scalar multiplication by `highPart` can be fully or partially hardcoded, because the points (`p1` and `p3`) are constants and the scalar is always 4 bits.
- [ ] instead of recovering the `x` coordinate in `Pedersen()` you can recover it in `PedersenArray()` and mutualise all the inverses `1/zi`

Will merge this now.

PR #345, opened 2023-02-16
feat: FFT signature now takes variadic options

gbotrel:
- feat: FFT signature now takes an option variadic
- feat: added fft.WithNbTasks
- feat: parallel.Execute with 1 task fast path


ThomasPiellard:
Could you just change the name `WithCoset` to `OnCoset` ?

PR #349, opened 2023-03-03
perf(stark-curve): no subgroup check on prime-order curve

yelhousni:


ivokub (approved):
Looks good.

I'm trying to understand what was the previous behaviour - it seems that the method didn't depend on `p` at all and always returned true?

For long-term, should we add tests which try to pass non-valid points and check that the result is false?

yelhousni:
> I'm trying to understand what was the previous behaviour - it seems that the method didn't depend on `p` at all and always returned true?
> 
Oh right! the `res` point was always the zero point which is always on curve and has `z=0` in Jacobian coordinates. So the tests pass no matter what `p` is.. it should have been `res.ScalarMultiplication(p, fr.Modulus())`. But since the curve has a prime order we just check if `p` is on the curve (as for bn254/g1 and secp256k1).
> For long-term, should we add tests which try to pass non-valid points and check that the result is false?
Some invalid test vectors here would definitely be useful !

PR #363, opened 2023-03-22
fix: invalid infinity point decoding throws error

gbotrel:
Thanks to @asanso and @kevaundray for reporting; 

IETF spec we followed states:

> The serialization of a point at infinity comprises a string of zero bytes, except that the metadata bits may be nonzero.[¶](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-11#appendix-C-3.3)

Previously, we did check the metadata bits and if they were signaling a point at infinity, discarded the rest of the buffer. We now add an additional check to ensure all the bytes are at 0.


yelhousni (approved):
LGTM

PR #370, opened 2023-04-05
Add `CopyWithFreshBuffer` a function to copy the state

AlexandreBelling:
### Adds a method `CopyWithFreshBuffer` to the `RSis` struct.

*Returns a clone of the RSis parameters with a fresh and empty buffer. Does not mutate the current instance. The keys and the public parameters of the SIS instance are not deep-copied. It is useful when we want to hash in parallel. Otherwise, we would have to generate an entire RSis for each thread.*


ivokub:
Looks good to me, but I don't have experience in this part of the code. @ThomasPiellard, @Tabaie, can you have a look?

PR #372, opened 2023-04-05
make `mapToCurve` public to allow for custom cofactor clearing

hussein-aitlahcen:
This PR allow third party software to rely on the `mapToCurve` provided by gnark-crypto independently from the cofactor clearing.


Tabaie (approved):
Thank you

ivokub:
This PR is against master. In general we base against develop first and then merge whole develop into master.

Without knowing the context, was it intentional?

hussein-aitlahcen:
> @Tabaie already confirmed that PR is good. Halting only until figuring out which branch to base on.

Oh, sorry, wasn't intentional, do I base against develop then?

Tabaie:
Develop should be good!

hussein-aitlahcen:
Updated!

ivokub (approved):
In that case, good to go!

PR #377, opened 2023-04-07
Fix/fold kzg fs

Tabaie:
When batching KZG openings, the Fiat-Shamir challenge must also depend on the claimed evaluation.


ThomasPiellard (approved):

sandritis2:
fix/fold-kzg-fs

PR #379, opened 2023-04-09
Feat/encode uint64 slices

Tabaie:
In the Plonk verification key, when multi-commits are implemented, we will have to store an array of indexes to commitment constraints. This PR enables that by storing slices of `uint64` in the `len(x), x[0], x[1], ..., x[len(x)-1]` format.

### DIFF
--- a/ecc/bls12-377/marshal.go
+++ b/ecc/bls12-377/marshal.go
@@ -103,6 +103,24 @@ func (dec *Decoder) Decode(v interface{}) (err error) {
 	var read int
 
 	switch t := v.(type) {
+	case *[]uint64:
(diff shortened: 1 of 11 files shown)


Tabaie:
> This one might not be necessary; the default
> 
> ```
> binary.Write(buf, binary.BigEndian, slice)
> ```
> 
> will catch the slide, and encode it without the length. The only issue is for decoding, you need to pass an initialized slice as argument. Does the PlonK verification key know how many indexes to read?

No the length of the array is how we know how many levels of commitments there are.

gbotrel (approved):

PR #380, opened 2023-04-12
Precompute point multiplication results in pedersen

omerfirmak:
Provides 5x performance improvement

### DIFF
--- a/ecc/stark-curve/pedersen-hash/pedersen_hash.go
+++ b/ecc/stark-curve/pedersen-hash/pedersen_hash.go
@@ -7,12 +7,12 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
+const nibbleCount = fp.Bits / 4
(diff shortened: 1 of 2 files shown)


omerfirmak:
@yelhousni, any chance we get an eye on this?

omerfirmak:
> > @yelhousni, any chance we get an eye on this?
> 
> So the point is instead of computing `highPart*p2 + lowPart * p1` you precompute a table of multiples of `p1` and `p2` that you later lookup?

Yeah, pretty much.

16 values are precomputed for each nibble in highPart and lowPart. Then according to values of each nibble one of those values is selected per nibble and they are all summed up for the end result.

yelhousni:
Right! I remember proposing something in these lines: https://github.com/ConsenSys/gnark-crypto/pull/327#issuecomment-1416053157

omerfirmak:
Thanks! Anything I can help with to get this merged?

PR #378, opened 2023-04-07
Refactor/break kzg srs

Tabaie:
The KZG SRS breaks up naturally into a prover component (G1[:]) and a verifier component (G1[0], G2[0:2]), which only have one G1 point in common (which is usually standardized and can be omitted anyway).
Treating it as a single object results in very long Plonk verifying keys for example, which can be problematic in Smart Contracts.

This PR implements the break-up.

WARNING: I've removed pretty much all pass-by-pointers I encountered. In case some of that was performance-critical, lmk and I'll add it back in.

PR #383, opened 2023-04-20
Refactor/breakup pedersen

Tabaie:
Similar to https://github.com/ConsenSys/gnark-crypto/pull/378 but for Pedersen commitments instead of KZG. When commitments are used, the verifier needs to check a proof of knowledge of commitment. However, only two G2 elements are required for this, and we should thus avoid storing a long slice of G1 elements inside the verifying key.

### DIFF
--- a/internal/generator/pedersen/template/pedersen.go.tmpl
+++ b/internal/generator/pedersen/template/pedersen.go.tmpl
@@ -2,62 +2,68 @@ import (
     "crypto/rand"
     "fmt"
     "github.com/consensys/gnark-crypto/ecc"
-    "github.com/consensys/gnark-crypto/ecc/{{.Name}}"
+    curve "github.com/consensys/gnark-crypto/ecc/{{.Name}}"
     "github.com/consensys/gnark-crypto/ecc/{{.Name}}/fr"
+	"io"
     "math/big"
 )
 
-// Key for proof and verification
-type Key struct {
-    g             {{.CurvePackage}}.G2Affine // TODO @tabaie: does this really have to be randomized?
-    gRootSigmaNeg {{.CurvePackage}}.G2Affine //gRootSigmaNeg = g^{-1/σ}
-    basis         []{{.CurvePackage}}.G1Affine
-    basisExpSigma []{{.CurvePackage}}.G1Affine
+// ProvingKey for committing and proofs of knowledge
+type ProvingKey struct {
+	basis         []curve.G1Affine
+	basisExpSigma []curve.G1Affine
 }
 
-func randomOnG2() ({{.CurvePackage}}.G2Affine, error) { // TODO: Add to G2.go?
-    gBytes := make([]byte, fr.Bytes)
-    if _, err := rand.Read(gBytes); err != nil {
-        return {{.CurvePackage}}.G2Affine{}, err
+type VerifyingKey struct {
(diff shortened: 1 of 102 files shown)


internal/generator/pedersen/template/pedersen.go.tmpl
```diff
@@ -2,62 +2,68 @@ import (
...
+}
+
+func randomFrSizedBytes() ([]byte, error) {
+    res := make([]byte, fr.Bytes)
+    _, err := rand.Read(res)
+    return res, err
+}
+
+func randomOnG2() (curve.G2Affine, error) { // TODO: Add to G2.go?
+	if gBytes, err := randomFrSizedBytes(); err == nil {
```
gbotrel, line 29: this is me nit picking but, the usual pattern we see in go code bases is 
if err != nil {return error} --> (else) return result.
(and not the opposite :p )

internal/generator/pedersen/template/pedersen.go.tmpl
```diff
@@ -93,3 +99,74 @@ func (k *Key) VerifyKnowledgeProof(commitment {{.CurvePackage}}.G1Affine, knowle
...
+// Marshal
+
+func (pk *ProvingKey) WriteTo(w io.Writer) (int64, error) {
+	enc := curve.NewEncoder(w)
+	//l := len(pk.basis)
+	if err := enc.Encode(uint32(len(pk.basis))); err != nil {
+		return enc.BytesWritten(), err
+	}
+
+	for i := range pk.basis {
```
gbotrel, line 113: can encode the slice directly (ie pass the slice as an argument to Encode) rather than a for loop on each element.

Same for decode.

internal/generator/pedersen/template/pedersen.test.go.tmpl
```diff
@@ -26,37 +26,43 @@ func randomFrSlice(t *testing.T, size int) []interface{} {
 	return res
 }
 
-func randomOnG1() ({{.CurvePackage}}.G1Affine, error) { // TODO: Add to G1.go?
-	gBytes := make([]byte, fr.Bytes)
-	if _, err := rand.Read(gBytes); err != nil {
-		return {{.CurvePackage}}.G1Affine{}, err
+func randomOnG1() (curve.G1Affine, error) { // TODO: Add to G1.go?
+	if gBytes, err := randomFrSizedBytes(); err == nil {
+		return curve.HashToG1(gBytes, []byte("random on g1"))
```
gbotrel, line 31: same remark; usual pattern is "if error -> return error"

internal/generator/pedersen/template/pedersen.go.tmpl
Tabaie: I thought since I was using both conditions it would be nice to consider the full half of the glass first for a change. 😁

internal/generator/tower/template/fq12over6over2/fq12.go.tmpl
```diff
@@ -209,15 +209,17 @@ func (z *E12) DecompressKarabina(x *E12) *E12 {
     var one E2
     one.SetOne()
 
-	// g3 == 0
-	if x.C1.B2.IsZero() {
+	if x.C1.B2.IsZero() /* g3 == 0 */ {
```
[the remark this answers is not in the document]
Tabaie: I must have merged in something from an experimental branch. Will try to "unbase".

PR #385, opened 2023-04-21
feat: fix v computation in ECDSA signature

ivokub:
I referred to other implementations to understand how v is computed when creating an ECDSA signature. I had assumed that v gives if `y` coordinate of the ephemeral point is small or large (e.h. bigger than half the order). But due to different encoding order, it was actually the oddity of `y`.

This is confirmed also in [Ethereum Yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf) (Appendix F) and makes also a lot more sense.


yelhousni (approved):
Ok makes sense!
image

PR #386, opened 2023-04-22
Perf: KZG verification in a single point

yelhousni:
This PR tradeoffs a scalar multiplication in **G2** for a scalar multiplication in **G1** in the case of KZG opening in a single point (`Verify` method). Kudos to @Tabaie for this observation.
```diff
-       // e([f(α) - f(a)]G₁, G₂).e([-H(α)]G₁, [α-a]G₂) ==? 1
+       // e([f(α)-f(a)+aH(α)]G₁], G₂).e([-H(α)]G₁, [α]G₂) == 1
```
We actually use KZG opening in multiple points (`BatchVerifyMultiPoints`) for PlonK. So this does not impact the PlonK verification time but only the public `Verify` method, which could be of independent interest. On AWS z1d.large:

```go
benchmark                       old ns/op     new ns/op     delta
BenchmarkKZGVerify-2            708830        653944        -7.74%
BenchmarkKZGBatchVerify10-2     1084355       1035049       -4.55%
```

### DIFF
--- a/ecc/bls12-377/fr/kzg/kzg.go
+++ b/ecc/bls12-377/fr/kzg/kzg.go
@@ -195,28 +195,19 @@ func Verify(commitment *Digest, proof *OpeningProof, point fr.Element, vk Verify
 	var negH bls12377.G1Affine
 	negH.Neg(&proof.H)
 
-	// [α-a]G₂
-	var alphaMinusaG2Jac, genG2Jac, alphaG2Jac bls12377.G2Jac
+	// [f(α) - f(a) + a*H(α)]G₁
+	var totalG1 bls12377.G1Jac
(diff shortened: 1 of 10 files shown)


Tabaie (approved):
LGTM! Just a couple of questions:
1. Where does the improved performance come from? That scalar multiplication is quicker for G1 than for G2?
2. Is there anything to be gained from changing `e([f(α)-f(a)+aH(α)]G₁], G₂).e([-H(α)]G₁, [α]G₂) == 1` to `e([f(α)-f(a)+aH(α)]G₁], G₂) == e([H(α)]G₁, [α]G₂)` and avoiding a GT multiplication or is that relatively trivial?

PR #390, opened 2023-05-01
Generic slice serialization

Tabaie:
This PR enables the user to serialize (potentially multi-dimensional) slices through the `encoder.Encode` and `decoder.Decode` functions. It is a last resort so will not interfere with custom implementations in performance-critical applications as this implementation uses reflection heavily (per element).


Tabaie:
Abandoned over performance concerns

PR #395, opened 2023-05-04
Fix some implicit memory aliasing in for loops

jtraglia:
This PR fixes some implicit memory aliasing in for loops, which weren't being caught by `golangci-lint` (`gosec`).

It should probably be noted that golangci-lint does not run on generated files.


gbotrel:
> It should probably be noted that golangci-lint does not run on generated files.
!!! really? 😮

jtraglia:
Yes, I can't find a good document that states this, but [here's the code](https://github.com/golangci/golangci-lint/blob/7dcc1a1a6a0a9f6fa248eb225f15eeebe2d15fd6/pkg/result/processors/autogenerated_exclude.go#L72-L80):

```go
// isGenerated reports whether the source file is generated code.
// Using a bit laxer rules than https://go.dev/s/generatedcode to
// match more generated code. See #48 and #72.
func isGeneratedFileByComment(doc string) bool {
	const (
		genCodeGenerated = "code generated"
		genDoNotEdit     = "do not edit"
		genAutoFile      = "autogenerated file" // easyjson
	)
```

If the file contains "do not edit", which the generated files do, it will skip it.

jtraglia:
And sorry, I realized that I changed the generated files instead of the generator, and that there's probably more than three instances of this problem. It's going take a little while to do it right. Marking it as a draft in the mean time.

jtraglia:
So, I believe this is fixed now. Alternatively, we could replace this:

```go
for _, s := range testValues {
    s := s // prevent memory aliasing
    myFunc(&s)
}
```

with something like:

```go
for i := range testValues {
    myFunc(&testValues[i])
}
```

I think I prefer the second option, but the first option requires fewer changes.

gbotrel (approved):

PR #396, opened 2023-05-04
Run golangci-lint on generated files

jtraglia:
As mentioned in #395, `golangci-lint` does not run on generated files. This PR adds a hack that removes the "generated by" comments before running `golangci-lint` in the CI staticcheck test. This should trick it into examining them. I expect this PR's CI checks to fail until these findings are fixed, probably in other PRs.


gbotrel:
nice hack, will fix in separate PR 👍

gbotrel (approved):

PR #397, opened 2023-05-05
Disable check shadowing in govet linter

jtraglia:
This PR removes the `linter-settings` block which enables `check-shadowing`. Personally, I don't find this check all that useful; it's not necessarily dangerous and making new variables like `err2` and `err3` is pretty ugly IMO.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,7 +1,3 @@
-linters-settings:
-  govet:
-    check-shadowing: true
-
 linters:
   disable-all: true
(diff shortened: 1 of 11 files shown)


jtraglia:
Also worth mentioning, the `.golangci.yml` file in `gnark` doesn't have this either:

* https://github.com/ConsenSys/gnark/blob/master/.golangci.yml

gbotrel (approved):

PR #406, opened 2023-05-23
docs: ConsenSys -> Consensys

Tabaie:
Company name has changed from "ConsenSys" to "Consensys". This PR reflects that change in the copyright notice.

### DIFF
--- a/ecc/bls12-377/ecdsa/doc.go
+++ b/ecc/bls12-377/ecdsa/doc.go
@@ -1,4 +1,4 @@
-// Copyright 2020 ConsenSys Software Inc.
+// Copyright 2020 Consensys Software Inc.
 //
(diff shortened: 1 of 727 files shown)


ecc/bls12-377/ecdsa/doc.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 ConsenSys Software Inc.
+// Copyright 2020 Consensys Software Inc.
```
Tabaie, line 1: Should this change to 2023 as well or is it supposed to be the original publishing date?

gbotrel: normally, it should reflect the initial creation date.

gbotrel (approved):

PR #407, opened 2023-05-24
Feat/fold pedersen

Tabaie:
In anticipation of multicommits for Groth16, this PR provides the possibility to verify proofs of knowledge for Pedersen commitments more efficiently (by reducing the number of MSMs, generating the proof might be a bit faster as well)

Similarly to KZG, a value $r$ is obtained from a single-challenge Fiat-Shamir object using byte arrays provided by the user (the reason for custom byte arrays explained later.) Then instead of shipping $\set{(C_i, \pi_i)\}_i$ where $C_i$ are commitments and $\pi_i$ the corresponding proofs of knowledge, the prover sends $(\set{C_i}_i, \pi \coloneqq \sum_i r^i\pi_i)$.
The verifier in turn computes $C \coloneqq \sum_i r^iC_i$ and instead of checking $e(C_i,g)=e(\pi_i,g^{1/\sigma})$ multiple times, checks once that $e(C,g)=e(\pi,g^{1/\sigma})$.

From the user's perspective, the breaking change is that the function $pk.Commit$ does not yield a proof of knowledge anymore. A later call to $pk.ProveKnoweldge$ or $BatchProve$ will be necessary.

**Remark:** Normally in Fiat-Shamir one would expect hashing all the commitments that are to be verified, perhaps along with a nonce coming from the user. However, if we allow the user to take control of this, in the Groth16+Commitments verifier we could hash the representations of the commitment values in Fr, rather than the larger group elements. An alternative could be for the `Fold` functions to hash the compressed representations of the commitment values directly. Would welcome your input on this design choice.

### DIFF
--- a/ecc/bls12-377/fr/pedersen/pedersen.go
+++ b/ecc/bls12-377/fr/pedersen/pedersen.go
@@ -18,10 +18,12 @@ package pedersen
 
 import (
 	"crypto/rand"
+	"crypto/sha256"
 	"fmt"
 	"github.com/consensys/gnark-crypto/ecc"
 	curve "github.com/consensys/gnark-crypto/ecc/bls12-377"
 	"github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
+	fiatshamir "github.com/consensys/gnark-crypto/fiat-shamir"
 	"io"
 	"math/big"
 )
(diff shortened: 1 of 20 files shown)


ThomasPiellard:
lgtm, just to be sure could you add a test where a proof is tampered to check that it "fails successfully ^^" ?

PR #411, opened 2023-06-14
fix: incorrect semaphore init could cause msm deadlock

gbotrel:


ivokub (approved):
LGTM. Also tested PR with previously failing edge case and it is now fixed.

PR #435, opened 2023-07-23
Small optimization over the memory usage of MiMC

AlexandreBelling:
## Description

- `Reset` keeps the same buffer
- `tmp` is allocated once above the loop (turns out that the allocation is not optimized)


ivokub (approved):
Looks good. I'm still a bit puzzled that Go didn't optimize the tmp var allocation inside loop.
I tried a very minimal generic version wihout asm methods of encrypt method in godbolt, but it didn't show any difference in the compiled code. But I didn't use the assembly -optimized squaring, so maybe Go cannot optimize then.

PR #437, opened 2023-08-16
perf: improve fft domain  memory footprint

gbotrel:
This PR removes precompute of `CosetTableReversed` and `CosetTableInvReversed` in `fft.Domain`. 
In practice these can be accessed from `CosetTable` and `CosetTableInv` .

New version may result in slightly slower FFT on small domains. 

However, for large domains (think large plonk circuit), precomputing these and storing them in RAM contribute to a high memory load, hence the PR.


yelhousni (approved):
This makes sense for large zkEVM circuits indeed. What is the exact slowdown on small circuits, say a 2-chain Groth16 verifier?

gbotrel:
Benchmarked `groth16_bls12377` , virtually no delta on the M1 👍

gbotrel:
1% slower for a larger circuit.

PR #441, opened 2023-09-04
edwards: improve the performance of Add, MixedAdd and IsOnCurve

jsign:
# Description

This PR improves the performance if Edwards-related curves in methods: `Add`, `MixedAdd` and `IsOnCuve`. The crux of the problem is two indirect allocations produced by `GetEdwardsCurve(...)` call.

I’m pretty sure that, at least for `MixedAdd`, this is a regression since we didn’t have this allocation some version ago. While fixing it, I noticed this was also happening today for `Add` and `IsOnCuve`, so I’m unsure if those are regressions.

The benchmarks also surface a performance problem I’m not fixing in this PR to avoid convoluting topics. I’ll give a brief explanation in the benchmark section below.

## Type of change

- [x]  Performance improvement

# How has this been tested?

This PR doesn’t change the logic or add new border cases, so existing tests cover correct behavior.

The three mentioned methods (`Add`, `MixedAdd` and `IsOnCurve`) didn’t have benchmarks for Edwards curves, so that’s potentially the reason this was not detected before. I’ve created new benchmarks for each of them to measure the effect of this change.

Running these benchmarks before and after this PR:

```jsx
$ benchstat pre.pprof post.pprof                                                             
name                    old time/op    new time/op    delta
MixedAdd/Projective-16     488ns ± 3%     218ns ± 0%   -55.23%  (p=0.000 n=10+8)
MixedAdd/Extended-16       221ns ± 2%     224ns ± 5%      ~     (p=0.148 n=10+10)
Add/Affine-16             4.86µs ± 1%    4.63µs ± 3%    -4.75%  (p=0.000 n=10+10)
Add/Projective-16          501ns ± 2%     238ns ± 1%   -52.45%  (p=0.000 n=10+8)
Add/Extended-16           3.70µs ± 1%    3.68µs ± 3%      ~     (p=0.255 n=10+10)
IsOnCurve/positive-16      357ns ± 3%     107ns ± 4%   -70.03%  (p=0.000 n=9+10)
IsOnCurve/negative-16      351ns ± 3%     108ns ± 4%   -69.12%  (p=0.000 n=9+10)

name                    old alloc/op   new alloc/op   delta
MixedAdd/Projective-16      128B ± 0%        0B       -100.00%  (p=0.000 n=10+10)
MixedAdd/Extended-16       0.00B          0.00B           ~     (all equal)
Add/Affine-16               128B ± 0%        0B       -100.00%  (p=0.000 n=10+10)
Add/Projective-16           128B ± 0%        0B       -100.00%  (p=0.000 n=10+10)
Add/Extended-16            0.00B          0.00B           ~     (all equal)
IsOnCurve/positive-16       128B ± 0%        0B       -100.00%  (p=0.000 n=10+10)
IsOnCurve/negative-16       128B ± 0%        0B       -100.00%  (p=0.000 n=10+10)

name                    old allocs/op  new allocs/op  delta
MixedAdd/Projective-16      2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
MixedAdd/Extended-16        0.00           0.00           ~     (all equal)
Add/Affine-16               2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
Add/Projective-16           2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
Add/Extended-16             0.00           0.00           ~     (all equal)
IsOnCurve/positive-16       2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
IsOnCurve/negative-16       2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

Note: you can run the “before” by checking out commit `53784cadc54d39c1d1` where I’ve included the benchmarks for `master` without the performance improvements.
Note-2: I'm only showing differences for Bandersnatch; but the same would apply to other edwards.

In summary: 2x speedup in computation, and removing both existing allocations.

Now, notice the `Add/Extended` benchmark. This result is unchanged, which makes sense since it didn’t call `GetEdwardsCurve(...)` in the implementation, but the performance is odd since it should be faster than projective. I know the reason, so I[’d recommend seeing this draft PR](https://github.com/Consensys/gnark-crypto/pull/442) and maybe continuing the discussion there. :)

### DIFF
--- a/internal/generator/edwards/template/point.go.tmpl
+++ b/internal/generator/edwards/template/point.go.tmpl
@@ -145,8 +145,7 @@ func NewPointAffine(x, y fr.Element) PointAffine {
 
 // IsOnCurve checks if a point is on the twisted Edwards curve
 func (p *PointAffine) IsOnCurve() bool {
-
-	ecurve := GetEdwardsCurve()
+	initOnce.Do(initCurveParams)
 
 	var lhs, rhs, tmp fr.Element
 
@@ -158,7 +157,7 @@ func (p *PointAffine) IsOnCurve() bool {
 	tmp.Mul(&p.X, &p.X).
 		Mul(&tmp, &p.Y).
 		Mul(&tmp, &p.Y).
-		Mul(&tmp, &ecurve.D)
+		Mul(&tmp, &curveParams.D)
 	rhs.SetOne().Add(&rhs, &tmp)
 
 	return lhs.Equal(&rhs)
@@ -174,8 +173,7 @@ func (p *PointAffine) Neg(p1 *PointAffine) *PointAffine {
 // Add adds two points (x,y), (u,v) on a twisted Edwards curve with parameters a, d
 // modifies p
 func (p *PointAffine) Add(p1, p2 *PointAffine) *PointAffine {
-
-	ecurve := GetEdwardsCurve()
+	initOnce.Do(initCurveParams)
 
 	var xu, yv, xv, yu, dxyuv, one, denx, deny fr.Element
 	pRes := new(PointAffine)
@@ -188,7 +186,7 @@ func (p *PointAffine) Add(p1, p2 *PointAffine) *PointAffine {
 	yv.Mul(&p1.Y, &p2.Y)
 	pRes.Y.Sub(&yv, &xu)
 
-	dxyuv.Mul(&xv, &yu).Mul(&dxyuv, &ecurve.D)
+	dxyuv.Mul(&xv, &yu).Mul(&dxyuv, &curveParams.D)
 	one.SetOne()
 	denx.Add(&one, &dxyuv)
 	deny.Sub(&one, &dxyuv)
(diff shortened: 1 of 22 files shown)


internal/generator/edwards/template/point.go.tmpl
```diff
@@ -145,8 +145,7 @@ func NewPointAffine(x, y fr.Element) PointAffine {
 
 // IsOnCurve checks if a point is on the twisted Edwards curve
 func (p *PointAffine) IsOnCurve() bool {
-
-	ecurve := GetEdwardsCurve()
+	initOnce.Do(initCurveParams)
```
jsign, line 148: Taking the same strategy as in other places.

internal/generator/edwards/template/tests/point.go.tmpl
```diff
@@ -800,4 +800,100 @@ func BenchmarkNeg(b *testing.B) {
...
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			_ = point.IsOnCurve()
+		}
+	})
+
+	b.Run("negative", func(b *testing.B) {
+		var point PointAffine
+		point.ScalarMultiplication(&params.Base, &s)
+		point.X.Add(&point.X, &point.X)
```
jsign, line 892: Forcing X to be out of the curve. I double-checked that `IsOnCurve` returns `false` in this case.

yelhousni: hmm ok it works, if you checked with all Base points on all curves. Anyway, if `(x,y)` is on curve then `-a*x^2+y^2 == 1 + d*x^2*y^2` but `-a*(2x)^2+y^2 != 1 + d*(2x)^2*y^2` because none of the Base points have `x=0` or `y^2=-a/d` I guess.

yelhousni (approved):
Thanks for the PR @jsign ! This looks good to me. We have not optimised a lot twisted Edwards as they are only used in tests on the gnark side. If you have more optimisations, keep them coming our way! Thanks!

internal/generator/edwards/template/tests/point.go.tmpl
jsign: Good point. I could add a further check before the benchmark look that tests that on-call returns `false`, and fail fast if that isn't the case.

That way we could confirm that what's being benchmarked is actually correct.

jsign: @yelhousni, pushed an extra commit with this if you want to take a look.

yelhousni: yeah better, if end up adding more curve 👍

PR #445, opened 2023-09-06
Perf: specialize the multiplication by D in twistededward packages

yelhousni:
# Description

This PR was motivated by #442 and #441. Twisted Edwards curves are not as optimized as Weierstrass ones in gnark-crypto, since the formers are only used for tests in gnark. Anyway, this PR is an example to push the performance a bit further. 

Multiplication by the constant `D` is used is several places:
- Affine addition
- projective addition
- projective mixed addition 
- extended addition
- Curve membership check
- `X` coordinate retrieval

This PR specialize this multiplication for baby-Jubjub, Jubjub and Bandersnatch twisted Edwards curves. The same can be done for other curves but only these are actually used in projects (maybe BLS12-377 companion curve too e.g. Aleo, Penumbra?).

## Type of change

- [x] Performance improvement

# How has this been tested?

This PR doesn’t change the logic or add new border cases, so existing tests cover correct behavior.

- For baby-Jubjub, this saves in `Fr`: 5 doublings and 81 additions;
- For Jubjub, this saves in `Fr`: 6 doublings and 79 additions;
- For Bandersnatch, this saves in `Fr`: 6 doublings and 78 additions.


yelhousni:
> So maybe something is odd?
Oh wow!! ^^ that's so weird, let me check

yelhousni:
ok that was a silly PR, my bad :) In my head, it was a generic scalar multiplication that I replaced by a short addition chain but here the multiplication is in `Fp` so that's really nonsense :) Closing this.

PR #450, opened 2023-09-22
Improve performance of point equality checks

jsign:
# Description

This PR optimizes how point equality checks are done for G1 and twedwards. The previous code transformed points to affine points and then did equality checks requiring two inverses. The new code avoids doing inverses. (I wonder if there was a reason for the current implementation, or if I might need to include something, LMK!).

I also added benchmark codes so we can compare the old and new code.

## Type of change

- [X] Performance improvement

# How has this been tested?

It runs all existing tests. If this equality check code change breaks the correct logic, it will probably make many tests fail.

I created new benchmarks for equality checks. To test it against `master`, I backported them to the current `master`.

This was run on `AMD Ryzen 7 3800XT 8-Core Processor`.

For succinctness, I'll show the performance difference between BLS12-381 and Bandersnatch curves (but the implementation should fix all the other curves too).

Bandersnatch:
```
name                    old time/op  new time/op  delta
ProjEqual/equal-16      2.45µs ± 2%  0.07µs ± 1%  -97.26%  (p=0.000 n=10+9)
ProjEqual/not_equal-16  2.39µs ± 2%  0.04µs ± 1%  -98.51%  (p=0.000 n=10+9)
```

BLS12-381:
```
name                     old time/op  new time/op  delta
G1JacEqual/equal-16      4.95µs ± 3%  0.23µs ± 1%  -95.44%  (p=0.000 n=10+8)
G1JacEqual/not_equal-16  4.79µs ± 5%  0.11µs ± 1%  -97.61%  (p=0.000 n=9+9)
``` 

In both cases, we can see that `not_equal` bench is faster than `equal` which makes sense since we compare `X` first and `Y` afterwards. In the old code, most of the time is dominated by inverses so `equal` and `not_equal` have similar-ish perf.

### DIFF
--- a/internal/generator/ecc/template/point.go.tmpl
+++ b/internal/generator/ecc/template/point.go.tmpl
@@ -207,9 +207,14 @@ func (p *{{ $TJacobian }}) Set(a *{{ $TJacobian }}) *{{ $TJacobian }} {
 
 // Equal tests if two points (in Jacobian coordinates) are equal
 func (p *{{ $TJacobian }}) Equal(a *{{ $TJacobian }}) bool {
-
-	if p.Z.IsZero() && a.Z.IsZero() {
-		return true
+	// If one point is infinity, the other must also be infinity.
+	if p.Z.IsZero() {
+		return a.Z.IsZero()
+	}
+	// If the other point is infinity, return false since we can't
+	// the following checks would be incorrect.
+	if a.Z.IsZero() {
+		return false
 	}
 
 	var pZSquare, aZSquare {{.CoordType}}
--- a/internal/generator/ecc/template/point.go.tmpl
+++ b/internal/generator/ecc/template/point.go.tmpl
@@ -211,13 +211,21 @@ func (p *{{ $TJacobian }}) Equal(a *{{ $TJacobian }}) bool {
 	if p.Z.IsZero() && a.Z.IsZero() {
 		return true
 	}
-	_p := {{ $TAffine }}{}
-	_p.FromJacobian(p)
 
-	_a := {{ $TAffine }}{}
-	_a.FromJacobian(a)
+	var pZSquare, aZSquare {{.CoordType}}
+	pZSquare.Square(&p.Z)
+	aZSquare.Square(&a.Z)
 
-	return _p.X.Equal(&_a.X) && _p.Y.Equal(&_a.Y)
+	var lhs, rhs {{.CoordType}}
+	lhs.Mul(&p.X, &aZSquare)
+	rhs.Mul(&a.X, &pZSquare)
+	if !lhs.Equal(&rhs) {
+		return false
+	}
+	lhs.Mul(&p.Y, &aZSquare).Mul(&lhs, &a.Z)
+	rhs.Mul(&a.Y, &pZSquare).Mul(&rhs, &p.Z)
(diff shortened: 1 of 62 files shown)


internal/generator/ecc/template/point.go.tmpl
```diff
@@ -211,13 +211,21 @@ func (p *{{ $TJacobian }}) Equal(a *{{ $TJacobian }}) bool {
...
+	lhs.Mul(&p.X, &aZSquare)
+	rhs.Mul(&a.X, &pZSquare)
+	if !lhs.Equal(&rhs) {
+		return false
+	}
+	lhs.Mul(&p.Y, &aZSquare).Mul(&lhs, &a.Z)
+	rhs.Mul(&a.Y, &pZSquare).Mul(&rhs, &p.Z)
 
-	return _p.X.Equal(&_a.X) && _p.Y.Equal(&_a.Y)
+	return lhs.Equal(&rhs)
```
jsign, line 228: In short, we do `x_1*z_2^2 == x_2*z_1^2 && y_1 * z_2^3 == y_2 * z1^3`

internal/generator/edwards/template/point.go.tmpl
```diff
@@ -281,10 +281,17 @@ func (p *PointProj) Equal(p1 *PointProj) bool {
...
-	return pAffine.Equal(&p1Affine)
+
+	var lhs, rhs fr.Element
+	lhs.Mul(&p.X, &p1.Z)
+	rhs.Mul(&p1.X, &p.Z)
+	if !lhs.Equal(&rhs) {
+		return false
+	}
+	lhs.Mul(&p.Y, &p1.Z)
+	rhs.Mul(&p1.Y, &p.Z)
```
jsign, line 292: And here, `x_1*z_2 == x_2*z_1 && y_1 * z_2 == y_2 * z`

internal/generator/ecc/template/tests/point.go.tmpl
```diff
@@ -584,6 +584,51 @@ func Benchmark{{ $TJacobian }}IsInSubGroup(b *testing.B) {
...
+
+		aZScaled := a
+		aZScaled.X.Mul(&aZScaled.X, &scalarSquared)
+		aZScaled.Y.Mul(&aZScaled.Y, &scalarSquared).Mul(&aZScaled.Y, &scalar)
+		aZScaled.Z.Mul(&aZScaled.Z, &scalar)
+
+		// Check the setup.
+		if !a.Equal(&aZScaled) {
+			b.Fatalf("invalid test setup")
+		}
```
jsign, line 608: Building an equivalent point in jac coordinates, and double-checking that we'll benchmark what we think we'll benchmark.

internal/generator/edwards/template/tests/point.go.tmpl
```diff
@@ -732,6 +732,52 @@ func GenBigInt() gopter.Gen {
 // ------------------------------------------------------------
 // benches
 
+func BenchmarkProjEqual(b *testing.B) {
+	params := GetEdwardsCurve()
```
jsign, line 736: Similar strategy as done in Jac for Proj.

ThomasPiellard (approved):

PR #451, opened 2023-09-29
Fix: GLV exponentiation in GT with large exponents

ThomasPiellard:
# Description

* ExpGLV on 𝔽_{q}* /(𝔽_{q}*)ʳ uses max(len(s1), len(s2)) in the main loop (as in G1 and G2)
* mulWindowed failed for negative exponents ( [-s]G returned [s]G), the PR also fixes this.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

* Added a test that was previously failing with a big exponent for ExpGLV
* Added a test to check the property [-s]G=-[s]G

# Checklist:


yelhousni (approved):
All looks good 👍 
The last two commits just fix few typos and some cosmetics:
- merges the new (exp, expGLV) with the old (exp, expGLV and CyclotomicExp) test;
- uses `genA` variable to switch `GenE12()` for `GenE24()` or `GenE6()` when needed;
- uses `{{ toUpper .Name}}` for curves' names instead of BN254 in the new test;
- manually apply the fix (and test) to BLS24-* and BW6-* as the towers are not auto-generated there.

ecc/bls12-377/g1.go
```diff
@@ -421,8 +428,15 @@ func (p *G1Jac) mulWindowed(a *G1Jac, s *big.Int) *G1Jac {
 	var res G1Jac
 	var ops [3]G1Jac
 
-	res.Set(&g1Infinity)
 	ops[0].Set(a)
+	e := s
```
gbotrel, line 432: @yelhousni @ThomasPiellard  this snippet is weird?  didn't review this PR but to me this code is useless.
What does this code do; assign a *big.Int, overwrites it with an object from a pool, and Neg(s) into e, the object of the pool before discarding it.

ecc/bls12-381/g1.go
```diff
@@ -421,8 +428,15 @@ func (p *G1Jac) mulWindowed(a *G1Jac, s *big.Int) *G1Jac {
...
-	res.Set(&g1Infinity)
 	ops[0].Set(a)
+	e := s
+	if s.Sign() == -1 {
+		e = bigIntPool.Get().(*big.Int)
+		defer bigIntPool.Put(e)
+		e.Neg(s)
+		ops[0].Neg(&ops[0])
+	}
+	res.Set(&g1Infinity)
```
holiman, line 439: Just randomly browsed this particular thing. Looks like a lot of code with no side-effect. `e.Neg(s)` set `e` as `-s`, but otherwise is side-effect-free. And that's the last we see of `e`. So this whole chunk can be replaced by 
```golang
	if s.Sign() == -1 {
		ops[0].Neg(&ops[0])
	}
	res.Set(&g1Infinity)
```
Curius to know if I'm wrong?

PR #465, opened 2023-10-19
Refactor(BW6): use revisited Ate pairing instead of Tate

yelhousni:
# Description

TL;DR: use Eq. (6') instead of (6) in https://hackmd.io/@gnark/BW6-761-changes for BW6 Miller loops. This PR allows to test gnark in-circuit version (for BW6-761) which is needed for fixed-argument pairing. This would also be useful for all BW6 curves when we revisit KZG verification in gnark-crypto.

TODO: 
- [x] BW6-761
- [x] BW6-756
- [ ] BW6-633

## Type of change

# How has this been tested?

Same existing test work fine.

- [x] Benchmark on z1d.large
```
goos: linux
goarch: amd64
pkg: github.com/consensys/gnark-crypto/ecc/bw6-756
```
1-3% perf regression but would allow fixed-argument pairing which would largely offset this regression.

- BW6-761
```
benchmark                         old ns/op     new ns/op     delta
BenchmarkPairing-2                2099734       2135481       +1.70%
BenchmarkMultiPair/2_pairs-2      2869919       2924911       +1.92%
BenchmarkMultiPair/3_pairs-2      3627440       3712315       +2.34%
BenchmarkMultiPair/4_pairs-2      4387886       4499886       +2.55%
BenchmarkMultiPair/5_pairs-2      5144183       5285163       +2.74%
BenchmarkMultiPair/6_pairs-2      5905175       6076648       +2.90%
BenchmarkMultiPair/7_pairs-2      6656923       6846301       +2.84%
BenchmarkMultiPair/8_pairs-2      7410488       7631327       +2.98%
BenchmarkMultiPair/9_pairs-2      8167347       8415372       +3.04%
BenchmarkMultiPair/10_pairs-2     8919462       9196862       +3.11%
```

# Checklist:

### DIFF
--- a/internal/generator/ecc/template/point.go.tmpl
+++ b/internal/generator/ecc/template/point.go.tmpl
@@ -1521,52 +1521,6 @@ func (p *{{ $TProjective }}) FromAffine(Q *{{ $TAffine }}) *{{ $TProjective }} {
 	return p
 }
 
-{{- if eq .PointName "g1"}}
-// BatchProjectiveToAffine{{ toUpper .PointName }} converts points in Projective coordinates to Affine coordinates
-// performing a single field inversion (Montgomery batch inversion trick).
-func BatchProjectiveToAffine{{ toUpper .PointName }}(points []{{ $TProjective }}) []{{ $TAffine }} {
-	result := make([]{{ $TAffine }}, len(points))
-	zeroes := make([]bool, len(points))
-	accumulator := fp.One()
-
-	// batch invert all points[].Z coordinates with Montgomery batch inversion trick
-	// (stores points[].Z^-1 in result[i].X to avoid allocating a slice of fr.Elements)
-	for i := 0; i < len(points); i++ {
-		if points[i].z.IsZero() {
-			zeroes[i] = true
-			continue
-		}
-		result[i].X = accumulator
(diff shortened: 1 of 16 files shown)


internal/generator/ecc/template/point.go.tmpl
```diff
@@ -1521,52 +1521,6 @@ func (p *{{ $TProjective }}) FromAffine(Q *{{ $TAffine }}) *{{ $TProjective }} {
 	return p
 }
```
ivokub, line 1523: Why are we removing the method? Didn't this allow for efficient MSM?

ivokub (approved):
LGTM. But why remove the G1 projective definition?

internal/generator/ecc/template/point.go.tmpl
yelhousni: Projective coordinates iirc are used only for pairings (G2 for ate pairing and G1 for Tate pairing). This allows to mutualise some computations between the G1/G2 arithmetic and the line computation. But for scalar multiplication we use Jacobian coordinates as they are faster and for MSM we used extended Jacobians and affine coordinates.

ivokub: I see, thanks!

PR #466, opened 2023-11-03
Perf: KZG with fixed-argument pairing in affine coordinates

yelhousni:
# Description

This PR refactors KZG to use fixed-argument pairings. This requires to change the SRS verifying key to store the precomputed lines instead of the points G2 and [a]G2. 

TODO: 
- [x] BN254
- [x] BLS12-381
- [x] BLS12-377
- [x] BW6-761
- [x] BLS12-378
- [x] BW6-756
- [x] BLS24-317
- [x] BLS12-315
- [x] BW6-633

~I need to implement an ate variant Miller loop for BW6-633 to implement this feature.~ Done in [07b87ae](https://github.com/Consensys/gnark-crypto/pull/466/commits/07b87aea9e25c7ca3e62d417b7df8fd4fc9f22ae).

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How has this been tested?

Same tests for KZG work here as well + testing that `Pair` results with fixed and variable arguments match.

# How has this been benchmarked?

Some benchmarks on a AWS `z1d.large` machine:
```
goos: linux
goarch: amd64
pkg: github.com/consensys/gnark-crypto/ecc/*/kzg
cpu: Intel(R) Xeon(R) Platinum 8151 CPU @ 3.40GHz
```
- BN254
```
benchmark                          old ns/op      new ns/op      delta
BenchmarkSRSGen/real_SRS-2         3101757295     3085342398     -0.53%
BenchmarkSRSGen/quick_SRS-2        952900         1523358        +59.87%
BenchmarkKZGCommit/real_SRS-2      122914358      121965943      -0.77%
BenchmarkKZGCommit/quick_SRS-2     114308031      113640429      -0.58%
BenchmarkDivideByXMinusA-2         103447328      102627372      -0.79%
BenchmarkKZGOpen-2                 124670902      124335198      -0.27%
BenchmarkKZGVerify-2               658898         537151         -18.48%
BenchmarkKZGBatchOpen10-2          136281734      136385577      +0.08%
BenchmarkKZGBatchVerify10-2        1043165        922205         -11.60%
BenchmarkToLagrangeG1-2            5548069025     4977902867     -10.28%
```

- BW6-761
```
benchmark                          old ns/op       new ns/op       delta
BenchmarkSRSGen/real_SRS-2         26341215560     26358001933     +0.06%
[...]

PR #469, opened 2023-11-16
Refactor: re-introduce points in KZG verification key

yelhousni:
# Description

Re-introducing points `G1`, `G2` and `[α]G2` in KZG verification key after #466. This is needed in the Solidity Plonk verifier in gnark. My initial thought was to hardcode the values of the points' coordinates in solidity instead of calling the KZG package, but apparently (cc @ThomasPiellard) Linea (and Aztec) use a different `G1` point than `(1,2)` for BN254. We should also then use `ScalarMultiplication(x, G1)` instead of `ScalarMultiplicationBase(x)`. Anyway it should be flexible for users now.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue *on gnark side*)

# How has this been tested?

Same current tests.

NA

# Checklist:

### DIFF
--- a/ecc/bls12-377/kzg/kzg.go
+++ b/ecc/bls12-377/kzg/kzg.go
@@ -49,6 +49,8 @@ type ProvingKey struct {
 
 // VerifyingKey used to verify opening proofs
 type VerifyingKey struct {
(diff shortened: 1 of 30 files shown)


ivokub (approved):
LGTM

PR #471, opened 2023-12-06
perf: use inverse as a bijection for bw6-761 mimc

yelhousni:
# Description

Currently BW6-761 MiMC hash uses `pow5` as a bijection. This PR suggests to use the `Inverse` function instead. This reduces the number of constraints needed for the plonk verifier gadget in gnark (see https://github.com/Consensys/gnark/pull/949) but increases the out-circuit execution time.

TODO: 
- [ ] Security analysis of the number of rounds.

## Type of change

- [x] This change requires a documentation update

# How has this been tested?

Same tests as in the master branch.

Please describe the benchmarks that you ran to verify your changes.

- [ ] Benchmark A, on Macbook pro M1, 32GB RAM
- [ ] Benchmark B, on x86 Intel xxx, 16GB RAM

# Checklist:

### DIFF
--- a/ecc/bw6-761/fr/mimc/mimc.go
+++ b/ecc/bw6-761/fr/mimc/mimc.go
@@ -142,13 +142,9 @@ func (d *digest) checksum() fr.Element {
 func (d *digest) encrypt(m fr.Element) fr.Element {
 	once.Do(initConstants) // init constants
 
-	var tmp fr.Element
 	for i := 0; i < mimcNbRounds; i++ {
-		// m = (m+k+c)^5
-		tmp.Add(&m, &d.h).Add(&tmp, &mimcConstants[i])
-		m.Square(&tmp).
-			Square(&m).
-			Mul(&m, &tmp)
+		// m = 1/(m+k+c)
+		m.Add(&m, &d.h).Add(&m, &mimcConstants[i]).Inverse(&m)
 	}
 	m.Add(&m, &d.h)
 	return m
--- a/internal/generator/crypto/hash/mimc/template/mimc.go.tmpl
+++ b/internal/generator/crypto/hash/mimc/template/mimc.go.tmpl
@@ -177,6 +177,20 @@ func (d *digest) encrypt(m fr.Element) fr.Element {
 	m.Add(&m, &d.h)
 	return m
 }
+{{ else if eq .Name "bw6-761" }}
+// plain execution of a mimc run
+// m: message
+// k: encryption key
+func (d *digest) encrypt(m fr.Element) fr.Element {
+	once.Do(initConstants) // init constants
+
+	for i := 0; i < mimcNbRounds; i++ {
+		// m = 1/(m+k+c)
+		m.Add(&m, &d.h).Add(&m, &mimcConstants[i]).Inverse(&m)
+	}
+	m.Add(&m, &d.h)
+	return m
+}
 {{ else }}
 // plain execution of a mimc run
 // m: message


ivokub (commented):
Changes look good, but I'm not sure about the security analysis. Is there a reference which supports using inverse as a encryption function?

And in that case I'm not sure if we should keep it as MiMC or instead define a new hash function MiMC-Inverse?

@ThomasPiellard, do you have thoughts?

yelhousni:
> Changes look good, but I'm not sure about the security analysis. Is there a reference which supports using inverse as a encryption function?
> 
> And in that case I'm not sure if we should keep it as MiMC or instead define a new hash function MiMC-Inverse?
> 
> @ThomasPiellard, do you have thoughts?

Well we just need the s-box to be a bijection so given that `1/x == x^(p-2)` and `gcd(r-1,p-2)==1` then it should work. Now the security analysis part regarding the number of rounds is what I'm not sure about. For some reference, the inverse s-box is discussed in the Poseidon paper. But now that I am rereading the paper, I found that this dicussion was dismissed in the new updated version ([old version](https://eprint.iacr.org/archive/2019/458/1608125375.pdf) vs . [new version](https://eprint.iacr.org/2019/458.pdf)). 

From the Section "Previous Versions" in the Abstract of the new version, one can read:
> We emphasize that both (i) the entire family of STARKAD and (ii) POSEIDON instantiated with x → 1/x are considered as dismissed, and we advice against using them. 

I am not sure if it is fair to translate blindly the security analysis of Poseidon to MiMC but I think it would be safe to revert the use of the inverse s-box in MiMC. For the record, if we revert commit https://github.com/Consensys/gnark/pull/949/commits/bdab848e33f2305224d266fa1cf792c4bc40e1dd in PR https://github.com/Consensys/gnark/pull/949, the total nb of constraint is 342k instead of 317k for the 2-chain case (emulated case not impacted). So that's not terrible.

ivokub:
> > Changes look good, but I'm not sure about the security analysis. Is there a reference which supports using inverse as a encryption function?
> > And in that case I'm not sure if we should keep it as MiMC or instead define a new hash function MiMC-Inverse?
> > @ThomasPiellard, do you have thoughts?
> 
> Well we just need the s-box to be a bijection so given that `1/x == x^(p-2)` and `gcd(r-1,p-2)==1` then it should work. Now the security analysis part regarding the number of rounds is what I'm not sure about. For some reference, the inverse s-box is discussed in the Poseidon paper. But now that I am rereading the paper, I found that this dicussion was dismissed in the new updated version ([old version](https://eprint.iacr.org/archive/2019/458/1608125375.pdf) vs . [new version](https://eprint.iacr.org/2019/458.pdf)).
> 
> From the Section "Previous Versions" in the Abstract of the new version, one can read:
> 
> > We emphasize that both (i) the entire family of STARKAD and (ii) POSEIDON instantiated with x → 1/x are considered as dismissed, and we advice against using them.
> 
> I am not sure if it is fair to translate blindly the security analysis of Poseidon to MiMC but I think it would be safe to revert the use of the inverse s-box in MiMC. For the record, if we revert commit [Consensys/gnark@bdab848](https://github.com/Consensys/gnark/commit/bdab848e33f2305224d266fa1cf792c4bc40e1dd) in PR [Consensys/gnark#949](https://github.com/Consensys/gnark/pull/949), the total nb of constraint is 342k instead of 317k for the 2-chain case (emulated case not impacted). So that's not terrible.

Yup, skimming the different versions I think I have the same impression. I think for safety now we can revert bdab848 in 949 (to use pow5 instead of inv). With https://github.com/Consensys/gnark/pull/960 we get additional savings when batching and I think we already can fit many proofs in 2-chain currently.

I think we can save quite a lot more when going to 8- or 16-bit decomposition when marshalling G1 points. Right now the overhead is quite significant. And we can still look into implementing Pedersen instead of using MiMC.

But just in case I would leave this PR open. It may still work out after reconsidering nb of rounds.

PR #472, opened 2023-12-14
perf,memory: fft without precompute

gbotrel:
In short, this PR allows for `fft.NewDomain()` to have an option `fft.WithoutPrecompute()` which initializes the domain without precomputing (all) twiddles & coset tables.

For large domains, this save a huge amount of memory at the cost of 20-30% CPU perf downgrade. 

To simplify code, it modifies existing inner DIF and DIT fft methods to take a precomputed twiddle table and a starting stage; so twiddles can still be precomputed but we skip the first stages (each stage we skip we save domain.Cardinality / stage*2 memory).


ivokub (approved):
Conditional accept - I do not have enough knowledge about FFTs and twiddles to understand if done correctly. But making the computation optional seems correct.

If not urgent then I would ask @ThomasPiellard to confirm.

PR #474, opened 2023-12-19
feat: allow hashing Fiat-Shamir transcript challenge without hashing to field first

### LINKED ISSUE #467: refactor: make applying domain separation optional in Fiat-Shamir Transcript
**Is your feature request related to a problem? Please describe.**

Currently when the hashing function implements `WriteString`, then `fiatshamir.Transcript` uses this instead of `Write`. For MiMC implementation the `WriteString` prepends a domain separation string which is hardcoded. This implementation makes it difficult to be compatible with other implementations as this domain separation is not documented nor configurable.

**Describe the solution you'd like**

Instead on switching on `interface {WriteString(input string) (int, error)}` allow an option when initializing `fiatshamir.Transcript`. The syntax should be similar as in gnark. See https://github.com/Consensys/gnark/pull/900.

**Describe alternatives you've considered**
If user wants to have different domain separation processing, then they should wrap native hash functions to modify `WriteString` method. But in that case we still need to document the behaviour better.

cc @Tabaie - this would impact GKR and sumcheck, but it is already mitigated on gnark side.
cc @ThomasPiellard for visibility.

ivokub:
# Description

Fixes #467.

This PR allows to avoid hashing to field before we hash the challenge name into the transcript. The problem with the current approach was that it wasn't that well compatible with gnark counterpart when doing recursion in-circuit.

The obvious solution was to add options to conditionally allow altering the behaviour of Fiat-Shamir instance, but this created another problem as currently the challenge names were already given as variadic arguments and we cannot have multiple.

So, I removed mandatory challenge name fixing when initializing transcript and added an option to optionally fix them. Now, if the challenge names are not given then we can bind to new challenges on the go.

This allowed to add the option for avoiding the hash to field before hashing the challenge name into the transcript. Added edge case check where if hash to field is requested but we do not have the method.

This fix breaks usage of Fiat-Shamir transcript, but it is fixed in gnark-crypto and I'm also pushing changes to gnark. See consensys/gnark#974

## Type of change

# How has this been tested?

Tests work. Also checked that fixes the issue in zkevm.

Not benchmarked.

# Checklist:


ivokub:
Seems there is some incompatibility with gnark still.

gbotrel (approved):

PR #478, opened 2024-01-23
doc: update hash to curve rfc ref and impl comment

### LINKED ISSUE #477: bug: possibly incorrect `DST_prime` in `ExpandMsgXmd`
Unless I am missing something, it look like instead of `DST_prime = I2OSP(len(DST), 1) ∥ DST`, we do `DST_prime = DST || I2OSP(len(DST), 1) ` in:
- https://github.com/Consensys/gnark-crypto/blob/eb75782795d2d3ce6172797615f8eada77a6a3dd/field/hash/hashutils.go#L37-L40
- https://github.com/Consensys/gnark-crypto/blob/eb75782795d2d3ce6172797615f8eada77a6a3dd/field/hash/hashutils.go#L53-L56
- https://github.com/Consensys/gnark-crypto/blob/eb75782795d2d3ce6172797615f8eada77a6a3dd/field/hash/hashutils.go#L77-L80

hussein-aitlahcen:
fixes #477

### DIFF
--- a/field/hash/hashutils.go
+++ b/field/hash/hashutils.go
@@ -6,8 +6,8 @@ import (
 )
 
 // ExpandMsgXmd expands msg to a slice of lenInBytes bytes.
-// https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-06#section-5
-// https://tools.ietf.org/html/rfc8017#section-4.1 (I2OSP/O2ISP)
+// https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xmd
+// https://datatracker.ietf.org/doc/html/rfc9380#name-utility-functions (I2OSP/O2ISP)
 func ExpandMsgXmd(msg, dst []byte, lenInBytes int) ([]byte, error) {
 
 	h := sha256.New()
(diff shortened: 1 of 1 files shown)


field/hash/hashutils.go
```diff
@@ -6,8 +6,8 @@ import (
 )
 
 // ExpandMsgXmd expands msg to a slice of lenInBytes bytes.
-// https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-06#section-5
-// https://tools.ietf.org/html/rfc8017#section-4.1 (I2OSP/O2ISP)
+// https://tools.ietf.org/doc/html/rfc9380#name-expand_message_xmd
```
ivokub, line 9: The links redirect to IETF home page. Maybe use https://www.rfc-editor.org/rfc/rfc9380.html instead?

ivokub (changes requested):

field/hash/hashutils.go
ivokub: Or use https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xmd

hussein-aitlahcen: sure

hussein-aitlahcen:
@ivokub I updated the changes

ivokub (approved):
Perfect! Thanks for the contribution!

PR #490, opened 2024-03-04
fix: gkr add gate evaluate

BeratOz01:
# Description
This PR addresses a bug in the `gkr` package related to the `Evaluate` function of the `AddGate`. The existing logic within the AddGate evaluation function had an error that caused incorrect addition behavior under certain input scenarios. In addition to fixing the bug, this PR includes related tests as well.

https://github.com/Consensys/gnark-crypto/blob/d72fcb379d3eb894fc18ece74a91f835294e1439/ecc/bn254/fr/gkr/gkr.go#L881-L894

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?
# Checklist:

- [X] I have performed a self-review of my code
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I did not modify files generated from templates
- [X] `golangci-lint` does not output errors locally
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/ecc/bls12-377/fr/gkr/gkr.go
+++ b/ecc/bls12-377/fr/gkr/gkr.go
@@ -909,7 +909,7 @@ func (g MulGate) Evaluate(x ...fr.Element) (res fr.Element) {
 	default:
 		res.Mul(&x[0], &x[1])
 		for i := 2; i < len(x); i++ {
-			res.Mul(&res, &x[2])
+			res.Mul(&res, &x[i])
 		}
 	}
 	return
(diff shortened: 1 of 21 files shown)


ivokub (changes requested):
Good catch, thanks! However, it seems the same issue is for the MulGate (we multply always `x[2]` when more than two inputs). Can you also fix it?

Also, would it be possible to add regressions tests, i.e. tests where the AddGate takes three inputs and MulGate takes three inputs?

ivokub:
> Also, would it be possible to add regressions tests, i.e. tests where the AddGate takes three inputs and MulGate takes three inputs?

Hmm, looking at it, I'm not sure it is worth adding regression tests -- the prover and verifier use the same definition of the gates and it is difficult to test regression (without defining the correct AddGate for the verifier).

ivokub (approved):
I implemented fix for MulGate directly in your branch.

Thanks for the contribution, I'll merge it currently as is. Let us know if there is anything else you notice!

PR #492, opened 2024-03-18
fix: fixes #491

### LINKED ISSUE #491: Generator of Fr*
**Is your feature request related to a problem? Please describe.**
In the fft package the method to set a generator of Fr* is not exposed and the comment is wrong e.g.
```

	// generator of the largest 2-adic subgroup

	domain.FrMultiplicativeGen.SetUint64(5)
```
for bn254 should be "generator of Fr* ". This generator might be needed e.g. for fflonk.

**Describe the solution you'd like**
Add a public method GenFrStar()

ThomasPiellard:
# Description

Adds a method GenFrStar() in domain.go to get a generator of Fr* (wrongly named "generator of largest two adic subgroup" before").

Fixes # 491

## Type of change

# Checklist:


internal/generator/fft/template/domain.go.tmpl
```diff
@@ -45,37 +45,41 @@ type Domain struct {
...
-func NewDomain(m uint64, opts ...DomainOption) *Domain {
-	opt := domainOptions(opts...)
-	domain := &Domain{}
-	x := ecc.NextPowerOfTwo(m)
-	domain.Cardinality = uint64(x)
-
+// GetGenFrStar returns a generator of Z/rZ*
+func GetGenFrStar() fr.Element {
+	var res fr.Element
 	// generator of the largest 2-adic subgroup
```
ivokub, line 51: PR says that the comment was wrong?

internal/generator/fft/template/domain.go.tmpl
```diff
@@ -45,37 +45,41 @@ type Domain struct {
...
-// NewDomain returns a subgroup with a power of 2 cardinality
-// cardinality >= m
-// shift: when specified, it's the element by which the set of root of unity is shifted.
-func NewDomain(m uint64, opts ...DomainOption) *Domain {
-	opt := domainOptions(opts...)
-	domain := &Domain{}
-	x := ecc.NextPowerOfTwo(m)
-	domain.Cardinality = uint64(x)
-
+// GetGenFrStar returns a generator of Z/rZ*
```
ivokub, line 48: Maybe just add a comment in the package doc what `r` is? When referring to package documentation then it is better to find instead of having to look at the implementation.

For "nothing-up-my-sleeve" transparency maybe should also add a comment about the property of the generators? Are we using smallest generators or do they have some other properties?

internal/generator/fft/template/domain.go.tmpl
```diff
@@ -45,37 +45,41 @@ type Domain struct {
...
+func GetGenFrStar() fr.Element {
```
ivokub, line 49: And I think the method name could be different. Imo we usually do not use the `Get...` prefix for returning something. Maybe something a la `SubgroupGenerator` or `GeneratorSubgroup` etc considering we already have method `Generator`.

ivokub (approved):
Looks good, but I'm not sure about the method name, I think it deviates from out current method naming pattern a bit.

internal/generator/fft/template/domain.go.tmpl
ThomasPiellard: The generators do not have any particular properties, they were found using Sage

PR #500, opened 2024-05-03
refactor: move test utils package into isolation to avoid every user to import test stuff

gbotrel:
Previous state forced any package that imported `utils` (meaning a lot...) to import test stuff as side effect, and for example forced binaries building on top of gnark/gnark-crypto to embed testify/...


ivokub (approved):
LGTM!

PR #501, opened 2024-05-03
perf: more elegant solution to mem dump using unsafe and generics

gbotrel:
- **refactor: expose pedersen basis slices**
- **feat: more elegant solution to mem dump using unsafe and generics**

Motivation; enable other gnark objects (like `groth16.ProvingKey`) to be written and read very fast (though not portable).  Limit code duplication between curves, fields, proof systems, etc.

* Adds `utils/unsafe` with `WriteSlice[S ~[]E, E any](w io.Writer, s S) error` ... ; 
* `kzg.UnsafeToBytes(...) []byte` is now `WriteDump(w io.Writer) error` 

🔥  Benchmarks for kzg SRS dump 🔥 --> (I mean using `unsafe` is a cheat, we just benchmark the non-unsafe method + real IO)

```
benchmark                                      old ns/op      new ns/op      delta
BenchmarkSerializeSRS/WriteTo-10               1227334792     1200767042     -2.16%
BenchmarkSerializeSRS/WriteRawTo-10            1291239625     1275224792     -1.24%
BenchmarkSerializeSRS/UnsafeToBytes-10         288414938      49002102       -83.01%
BenchmarkDeserializeSRS/UnsafeReadFrom-10      1589721583     1607936042     +1.15%
BenchmarkDeserializeSRS/UnsafeFromBytes-10     283540219      81233306       -71.35%

benchmark                                      old allocs     new allocs     delta
BenchmarkSerializeSRS/WriteTo-10               540            540            +0.00%
BenchmarkSerializeSRS/WriteRawTo-10            542            541            -0.18%
BenchmarkSerializeSRS/UnsafeToBytes-10         522            512            -1.92%
BenchmarkDeserializeSRS/UnsafeReadFrom-10      1046           1047           +0.10%
BenchmarkDeserializeSRS/UnsafeFromBytes-10     1029           1031           +0.19%

benchmark                                      old bytes      new bytes      delta
BenchmarkSerializeSRS/WriteTo-10               2147544696     2147544696     +0.00%
BenchmarkSerializeSRS/WriteRawTo-10            4295028640     4295028632     -0.00%
BenchmarkSerializeSRS/UnsafeToBytes-10         1610821246     67176202       -95.83%
BenchmarkDeserializeSRS/UnsafeReadFrom-10      1627598752     1627598848     +0.00%
BenchmarkDeserializeSRS/UnsafeFromBytes-10     1610820720     1610820744     +0.00%
```

### DIFF
--- a/utils/unsafe/dump_slice.go
+++ b/utils/unsafe/dump_slice.go
@@ -1,0 +1,98 @@
+package unsafe
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"io"
+	"unsafe"
+)
+
+// WriteSlice writes a slice of arbitrary objects to the writer.
+// Use with caution, as it writes the raw memory representation of the slice;
+// In particular you do not want to use this with slices that contain pointers.
+// This architecture dependent and will not work across different architectures
+// (e.g. 32 vs 64 bit, big endian vs little endian).
+func WriteSlice[S ~[]E, E any](w io.Writer, s S) error {
+	var e E
+	size := int(unsafe.Sizeof(e))
+	if err := binary.Write(w, binary.LittleEndian, uint64(len(s))); err != nil {
+		return err
+	}
+
+	data := unsafe.Slice((*byte)(unsafe.Pointer(&s[0])), size*len(s))
+	if _, err := w.Write(data); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ReadSlice reads a slice of arbitrary objects from the reader, written by WriteSlice.
+func ReadSlice[S ~[]E, E any](r io.Reader, maxElements ...int) (s S, read int, err error) {
+	var buf [8]byte
+	if _, err := io.ReadFull(r, buf[:]); err != nil {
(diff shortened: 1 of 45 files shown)


utils/unsafe/dump_slice.go
```diff
@@ -0,0 +1,98 @@
...
+// This architecture dependent and will not work across different architectures
+// (e.g. 32 vs 64 bit, big endian vs little endian).
+func WriteSlice[S ~[]E, E any](w io.Writer, s S) error {
+	var e E
+	size := int(unsafe.Sizeof(e))
+	if err := binary.Write(w, binary.LittleEndian, uint64(len(s))); err != nil {
+		return err
+	}
+
+	data := unsafe.Slice((*byte)(unsafe.Pointer(&s[0])), size*len(s))
```
ivokub, line 23: Add guard in case `len(s) == 0`.

ivokub (approved):
LGTM. Imo we can add a simple guard in case the slice to write is empty to have better errors, but I think normally shouldn't happen.

PR #506, opened 2024-06-05
perf(bn254): optimize affine pairing with doubleAndAdd

yelhousni:
# Description

In affine pairing (`MillerLoopFixedQ`), use `doubleAndAdd` to compute `2P+Q` as of `P+Q+P`. This optimizes the one-time line precomputation for KZG in gnark-crypto, but most importantly matches gnark in-circuit implementation. This is now important since after https://github.com/Consensys/gnark/pull/1143 we need to have the exact same lines so that the witness residue is the same.

## Type of change

# How has this been tested?

pairing tests pass.

In gnark-crypto, only the lines precompuation should be affected (faster) but it's just a one-time computation.

# Checklist:


ivokub (approved):
LGTM

PR #510, opened 2024-06-12
perf(ecc): faster affine Add

yelhousni:
# Description

Actually, we can optimize affine addition more than #509 with the assumption that both `Z1` and `Z2` are 1 in Jacobian coordinates.  This PR uses [these EFD formulas](https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html#addition-mmadd-2007-bl). Also #509 modifies the second input in the `Sub` method which this PR fixes. 
https://github.com/Consensys/gnark-crypto/blob/364bbceb6d746d027c0e3f78912e61a129b51b8c/internal/generator/ecc/template/point.go.tmpl#L128

## Type of change

# How has this been tested?

same tests as #509 

e.g. BLS12-381 G1 addition on a z1d.large AWS machine:
```
goos: linux
goarch: amd64
pkg: github.com/consensys/gnark-crypto/ecc/bls12-381
BenchmarkG1AffineAdd
BenchmarkG1AffineAdd-2   	  232734	      5135 ns/op
PASS
```
This is a -3.56% saving compared to #509 
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkG1AffineAdd-2     5135          4952          -3.56%
```

# Checklist:

### DIFF
--- a/ecc/bls12-377/g1.go
+++ b/ecc/bls12-377/g1.go
@@ -87,12 +87,49 @@ func (p *G1Affine) ScalarMultiplicationBase(s *big.Int) *G1Affine {
 }
 
 // Add adds two point in affine coordinates.
+// Jacobian addition with Z1=Z2=1
+// https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html#addition-mmadd-2007-bl
 func (p *G1Affine) Add(a, b *G1Affine) *G1Affine {
 	var q G1Jac
-	q.FromAffine(a)
-	q.AddMixed(b)
-	p.FromJacobian(&q)
-	return p
+	// a is infinity, return b
+	if a.IsInfinity() {
+		p.Set(b)
+		return p
+	}
+	// b is infinity, return a
+	if b.IsInfinity() {
+		p.Set(a)
+		return p
+	}
+	if a.X.Equal(&b.X) {
+		// if b == a, we double instead
+		if a.Y.Equal(&b.Y) {
+			q.DoubleMixed(a)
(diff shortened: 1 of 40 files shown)


gbotrel (changes requested):
Code looks correct and match the formulas, however, it is not tested (i.e. G1Affine.Add is not called in any test files, only in `fr/lookup` and `kzg` ). 

Can you add some basic tests / property based tests now that we have a different code path than `Jacobian`?

(general addition, doubling, adding of point at infinity, and addition resulting in point at infinity)

yelhousni:
> Code looks correct and match the formulas, however, it is not tested (i.e. G1Affine.Add is not called in any test files, only in `fr/lookup` and `kzg` ).
> 
> Can you add some basic tests / property based tests now that we have a different code path than `Jacobian`?
> 
> (general addition, doubling, adding of point at infinity, and addition resulting in point at infinity)

I think in `TestG*AffineOps` we call affine `Add`, `Sub`, `Double` and `ScalarMul` here for example:
https://github.com/Consensys/gnark-crypto/blob/3ba464d9dfe41bb88458da9f687d78cf9a5c4838/ecc/bls12-381/g1_test.go#L266

This test
https://github.com/Consensys/gnark-crypto/blob/3ba464d9dfe41bb88458da9f687d78cf9a5c4838/ecc/bls12-381/g1_test.go#L250
checks the edge case where inputs are equal in `Add` but I do agree that tests involving the point `(0,0)` are missing. Will add.

gbotrel (approved):

PR #517, opened 2024-07-01
Feat/pedersen batch verify

Tabaie:
# Description

This PR implements a `BatchVerify` function that can verify $n$ proofs of knowledge from keys with the same base $G_2$ point but possibly different $\sigma$ parameters using $n+1$ pairings instead of $2n$.

The existing `BatchProve` method which folds both commitments and proofs remains valid by the following extractor construction, but is not suitable to some applications.

Let V be able to provide convincing folded proofs of knowledge for $C_i = x_i B$ for $i\in \{1,2 \}, B\in G_1$ with high probability. Then, an extractor is able, by rewinding, to obtain two correct folded proofs of knowledge $\Pi, \Pi' \in G_1$ corresponding to distinct challenges $r,r'\in \mathbb F_r$ respectively. Then we have $e(C_1 + rC_2,G^\sigma)=e(\Pi,G), e(C_1+r'C_2,G^\sigma)=e(\Pi',G).$ Then we get $e\big((r-r')C_2, G^\sigma\big)=e(\Pi-\Pi',G)\Rightarrow e(C_2,G^\sigma)=e\big(\frac 1 {r-r'}(\Pi-\Pi'), G\big)$, thus obtaining a proof in the original scheme of knowledge of $C_2$. A proof of knowledge of $C_1$ can similarly can be extracted.

The new scheme requires the prover to send the $n$ individual proofs to the verifier. The verifier randomizes $r$ without sending it as a challenge to the prover. It folds the commitments into $C = C_1 + rC_2+...+r^{n-1}C_n$ and checks that $e(C, G)=e(\Pi_1,G^{\frac 1 {\sigma_1}})...e(\Pi_n, G^{\frac 1{\sigma_n}})$. Then, from $1=e(C_1^{\sigma_1}-\Pi_1,G)e^r(C_2^{\sigma_2}-\Pi_2,G)...e^{r^{n-1}}(C_n^{\sigma_n}-\Pi_n,G)$ one can deduce with high probability that $e(C_i,G^{\sigma_i})=e(\Pi_i,G)$ for $i=1..n$.
## Type of change

- [x] This change requires a documentation update

## Interface changes
* The `Setup` function no longer takes the bases in variadic form. This is to provide the option to the user to set the `G2` point themselves instead of having it randomized.
* The prove and fold functions no longer take "Fiat-Shamir options". Instead, they take in a challenge field element, implementing an interactive version of the protocol. The user is now responsible for generating a suitable challenge should they choose to apply the Fiat-Shamir transformation.
* A `BatchVerify` function is added to verify many proofs of knowledge with the same G2 point efficiently. _It may be confusing to users that `BatchVerify` and `BatchProve` do not correspond._

# How has this been tested?

- [ ] TestSemiFoldProofs

# Checklist:


ecc/bls24-315/fr/pedersen/pedersen.go
```diff
@@ -212,34 +210,59 @@ func FoldCommitments(commitments []curve.G1Affine, fiatshamirSeeds ...[]byte) (c
...
 }
 
-func getChallenge(fiatshamirSeeds [][]byte) (r fr.Element, err error) {
-	// incorporate user-provided seeds into the transcript
-	t := fiatshamir.NewTranscript(sha256.New(), "r")
-	for i := range fiatshamirSeeds {
-		if err = t.Bind("r", fiatshamirSeeds[i]); err != nil {
-			return
+// BatchVerify verifies n separately generated proofs of knowledge from different setup ceremonies, using n+1 pairings rather than 2n.
+func BatchVerify(vk []VerifyingKey, commitments []curve.G1Affine, pok []curve.G1Affine, challenge fr.Element) error {
```
ivokub, line 225: Change argument `challenge` to `combinationCoeff` and denote in method documentation that has to be random for security.

Also add, that the verification keys have to have same `G2` point.

ivokub: And you can make the `challenge`/`combinationCoeff` optional using an option. When it is not provided then we generate randomly inside the method.

ivokub: And, what do you think about renaming the method to `BatchVerifyMulti` indicating maybe better that we can use different verification keys (with same G2 though)

internal/generator/ecc/template/point.go.tmpl
```diff
@@ -1822,3 +1825,19 @@ func batchAdd{{ $TAffine }}[TP p{{ $TAffine }}, TPP pp{{ $TAffine }}, TC c{{ $TA
 		(*R)[j].Set(&rr)
 	}
 }
+
+{{ if eq .PointName "g2"}}
+func RandomOnG2() (G2Affine, error) {
```
ivokub, line 1830: Should add method documentation.

ivokub: And I guess this method can be generic for both G1/G2 Affine points? Imo it could be useful.

But pinging @yelhousni in case the approach for generating random point is good and we don't introduce any security issue.

ivokub (changes requested):
Looks good to me, but I have some ideas on making more cleaner and flexible to use.

I messed up the comments though - I added mostly to the generated code, but they go for the template.

internal/generator/ecc/template/point.go.tmpl
```diff
@@ -1822,3 +1825,19 @@ func batchAdd{{ $TAffine }}[TP p{{ $TAffine }}, TPP pp{{ $TAffine }}, TC c{{ $TA
...
+	if gBytes, err := randomFrSizedBytes(); err != nil {
+		return G2Affine{}, err
+	} else {
+		return HashToG2(gBytes, []byte("random on g2"))
```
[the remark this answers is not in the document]
Tabaie: The benefit of this method is that even the randomizer wouldn't know the relative discrete log of the randomized point relative to the canonical generator so there's less "toxic waste" to deal with. I _think_ it's more efficient too (not that it matters much here.)

And I actually think this whole scheme is just as secure if we just use the canonical G2 generator but am not totally sure.

ivokub (approved):
Looks good to me now. I paraphrased the method documentation so that it would look nicely on the documentation site. I also added package examples which should show on on the documentation site, so the users would know how to combine different methods (and for example how to generate bases for the setup etc.). Have a look at my changes and they seem good, then it is good to merge on my side 👍

Tabaie:
Perfect thank you for the modifications!

PR #541, opened 2024-09-19
fix comment of v for `SetBigInt`

zhiqiangxu:
We can only conclude `v >= 0` instead of `v > 0` from `v.Cmp(&zero) != -1`.


ivokub:
Thanks for the contribution! I also code generated the implementation for all curves. I'll merge to a branch which collects all documentation fixes which we merge before next release.

ivokub (approved):

PR #543, opened 2024-09-21
Feat: Eisenstein integers 

yelhousni:
# Description

This PR implements Eisenstein integers arithmetic, including the half-GCD algorithm.
This follows: 
- https://ethresear.ch/t/fake-glv-you-dont-need-an-efficient-endomorphism-to-implement-glv-like-scalar-multiplication-in-snark-circuits/20394/7,
- https://en.wikipedia.org/wiki/Eisenstein_integer and 
- https://github.com/yelhousni/fake-GLV.

## Type of change

# How has this been tested?

Tests are implemented in `eisenstein_test.go`.

Half-GCD of some random 128-bit Eisenstein integers on a z1d.large AWS machine:
```
goos: linux
goarch: amd64
pkg: github.com/consensys/gnark-crypto/field/eisenstein
BenchmarkHalfGCD-2   	
248566157222030422477351007620901975759+127180499814568938453986077818539212698*ω
332172497967381992638818630507579627203+168595014618012844916999503528379356776*ω
250917177940023814038197936340989040662+62542345892514041158477824749997954710*ω
318633110928069828602520176758759340623+128691427653754930750011514601704135735*ω
33830068183445111887751763414565085374+45267849578544034355196683977077448326*ω 
189270309561050775793081572721741878623+306301444086216412434750350679994937465*ω
    6859	    171191 ns/op
```

# Checklist:

### DIFF
--- a/ecc/eisenstein_test.go
+++ b/ecc/eisenstein_test.go
@@ -212,6 +212,36 @@ func TestEisensteinArithmetic(t *testing.T) {
 	properties.TestingRun(t, gopter.ConsoleReporter(false))
 }
 
+func TestEisensteinHalfGCD(t *testing.T) {
+
+	t.Parallel()
+	parameters := gopter.DefaultTestParameters()
+	if testing.Short() {
+		parameters.MinSuccessfulTests = nbFuzzShort
+	} else {
+		parameters.MinSuccessfulTests = nbFuzz
+	}
+
+	properties := gopter.NewProperties(parameters)
+
+	genE := GenComplexNumber()
+
+	properties.Property("half-GCD", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			res := HalfGCD(a, b)
+			var c, d ComplexNumber
+			c.Mul(b, res[1])
+			d.Mul(a, res[2])
+			d.Add(&c, &d)
+			return d.Equal(res[0])
+		},
+		genE,
+		genE,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
 // GenNumber generates a random integer
 func GenNumber() gopter.Gen {
 	return func(genParams *gopter.GenParameters) *gopter.GenResult {
--- a/ecc/eisenstein_test.go
+++ b/ecc/eisenstein_test.go
@@ -261,3 +261,18 @@ func GenComplexNumber() gopter.Gen {
 		return &ComplexNumber{A0: values[0].(big.Int), A1: values[1].(big.Int)}
 	})
 }
+
+// bench
+func BenchmarkHalfGCD(b *testing.B) {
+	var prime, _ = new(big.Int).SetString("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffed", 16) // 2^255 - 19
+	a0, _ := rand.Int(rand.Reader, prime)
+	a1, _ := rand.Int(rand.Reader, prime)
+	c0, _ := rand.Int(rand.Reader, prime)
+	c1, _ := rand.Int(rand.Reader, prime)
+	a := ComplexNumber{A0: *a0, A1: *a1}
+	c := ComplexNumber{A0: *c0, A1: *c1}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		HalfGCD(&a, &c)
+	}
+}
--- a/ecc/eisenstein_test.go
+++ b/ecc/eisenstein_test.go
@@ -1,0 +1,232 @@
+package ecc
+
+import (
+	"math/big"
+	"math/rand"
+	"testing"
+
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/prop"
+)
+
+const (
+	nbFuzzShort = 10
+	nbFuzz      = 50
+)
+
+func TestEisensteinReceiverIsOperand(t *testing.T) {
+
+	t.Parallel()
+	parameters := gopter.DefaultTestParameters()
+	if testing.Short() {
+		parameters.MinSuccessfulTests = nbFuzzShort
+	} else {
+		parameters.MinSuccessfulTests = nbFuzz
+	}
+
+	properties := gopter.NewProperties(parameters)
+
+	genE := GenComplexNumber()
+
+	properties.Property("Having the receiver as operand (addition) should output the same result", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c, d ComplexNumber
+			d.Set(a)
+			c.Add(a, b)
+			a.Add(a, b)
+			b.Add(&d, b)
+			return a.Equal(b) && a.Equal(&c) && b.Equal(&c)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("Having the receiver as operand (sub) should output the same result", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c, d ComplexNumber
+			d.Set(a)
+			c.Sub(a, b)
+			a.Sub(a, b)
+			b.Sub(&d, b)
+			return a.Equal(b) && a.Equal(&c) && b.Equal(&c)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("Having the receiver as operand (mul) should output the same result", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c, d ComplexNumber
+			d.Set(a)
+			c.Mul(a, b)
+			a.Mul(a, b)
+			b.Mul(&d, b)
+			return a.Equal(b) && a.Equal(&c) && b.Equal(&c)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("Having the receiver as operand (neg) should output the same result", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			b.Neg(a)
+			a.Neg(a)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("Having the receiver as operand (conjugate) should output the same result", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			b.Conjugate(a)
+			a.Conjugate(a)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+func TestEisensteinArithmetic(t *testing.T) {
+
+	t.Parallel()
+	parameters := gopter.DefaultTestParameters()
+	if testing.Short() {
+		parameters.MinSuccessfulTests = nbFuzzShort
+	} else {
+		parameters.MinSuccessfulTests = nbFuzz
+	}
+
+	properties := gopter.NewProperties(parameters)
+
+	genE := GenComplexNumber()
+
+	properties.Property("sub & add should leave an element invariant", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c ComplexNumber
+			c.Set(a)
+			c.Add(&c, b).Sub(&c, b)
+			return c.Equal(a)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("neg twice should leave an element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			b.Neg(a).Neg(&b)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("conj twice should leave an element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			b.Conjugate(a).Conjugate(&b)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("add zero should leave element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			zero := new(ComplexNumber)
+			b.Add(a, zero)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("mul by one should leave element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			one := &ComplexNumber{
+				*big.NewInt(1),
+				*big.NewInt(0),
+			}
+			b.Mul(a, one)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("add should be commutative", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c, d ComplexNumber
+			c.Add(a, b)
+			d.Add(b, a)
+			return c.Equal(&d)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("add should be assiocative", prop.ForAll(
+		func(a, b, c *ComplexNumber) bool {
+			var d, e ComplexNumber
+			d.Add(a, b).Add(&d, c)
+			e.Add(c, b).Add(&e, a)
+			return e.Equal(&d)
+		},
+		genE,
+		genE,
+		genE,
+	))
+
+	properties.Property("mul should be commutative", prop.ForAll(
+		func(a, b *ComplexNumber) bool {
+			var c, d ComplexNumber
+			c.Mul(a, b)
+			d.Mul(b, a)
+			return c.Equal(&d)
+		},
+		genE,
+		genE,
+	))
+
+	properties.Property("mul should be assiocative", prop.ForAll(
+		func(a, b, c *ComplexNumber) bool {
+			var d, e ComplexNumber
+			d.Mul(a, b).Mul(&d, c)
+			e.Mul(c, b).Mul(&e, a)
+			return e.Equal(&d)
+		},
+		genE,
+		genE,
+		genE,
+	))
+
+	properties.Property("norm should always be positive", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			return a.Norm().Sign() > 0
+		},
+		genE,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+// GenNumber generates a random integer
+func GenNumber() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
+		elmt := *big.NewInt(rand.Int63())
+		genResult := gopter.NewGenResult(elmt, gopter.NoShrinker)
+		return genResult
+	}
+}
+
+// GenComplexNumber generates a random integer
+func GenComplexNumber() gopter.Gen {
+	return gopter.CombineGens(
+		GenNumber(),
+		GenNumber(),
+	).Map(func(values []interface{}) *ComplexNumber {
+		return &ComplexNumber{A0: values[0].(big.Int), A1: values[1].(big.Int)}
+	})
+}
--- a/ecc/eisenstein_test.go
+++ b/ecc/eisenstein_test.go
@@ -1,8 +1,8 @@
 package ecc
 
 import (
+	"crypto/rand"
 	"math/big"
-	"math/rand"
 	"testing"
 
 	"github.com/leanovate/gopter"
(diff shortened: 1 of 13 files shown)


ecc/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+		genE,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+// GenNumber generates a random integer
+func GenNumber() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
+		var prime, _ = new(big.Int).SetString("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffed", 16) // 2^255 - 19
```
kevaundray, line 248: Is 2^255 - 19 arbitrarily set for tests?

yelhousni: Yeah just some « random » prime for tests

field/eisenstein/eisenstein.go
```diff
@@ -0,0 +1,155 @@
...
+	norm.Sub(
+		norm,
+		temp.Mul(&z.A0, &z.A1),
+	)
+	return norm
+}
+
+// Quo sets z to the quotient of x and y, and returns z.
+func (z *ComplexNumber) Quo(x, y *ComplexNumber) *ComplexNumber {
+	norm := y.Norm()
```
ivokub, line 116: Should we explicitly check that the norm isn't zero? Otherwise we get panic below when dividing by norm. One convention is that we return nil and it is the callers responsibility to check that the result is not nil.

field/eisenstein/eisenstein.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+	norm.Sub(
+		norm,
+		temp.Mul(&z.A0, &z.A1),
+	)
+	return norm
+}
+
+// Quo sets z to the quotient of x and y, and returns z.
+func (z *ComplexNumber) Quo(x, y *ComplexNumber) *ComplexNumber {
```
ivokub, line 115: I'm not understanding how it works, is there a reference?

field/eisenstein/eisenstein.go
```diff
@@ -0,0 +1,155 @@
...
+	norm := y.Norm()
+	z.Conjugate(y)
+	z.Mul(x, z)
+	z.A0.Div(&z.A0, norm)
+	z.A1.Div(&z.A1, norm)
+	return z
+}
```
ivokub, line 122: And I guess it would make sense to implement `QuoRem` method which returns both quotient and remainder? Imo it seems in GCD below we need both anyway.

field/eisenstein/eisenstein.go
```diff
@@ -0,0 +1,155 @@
...
+
+	aRun.Set(a)
+	bRun.Set(b)
+	u.SetOne()
+	v.SetZero()
+	u_.SetZero()
+	v_.SetOne()
+	nbits := a.Norm().BitLen() / 2
+
+	for aRun.Norm().BitLen() > nbits {
```
ivokub, line 138: I guess there shouldn't be any edge cases where the length of `aRun` doesn't decrease. But maybe we should have a check that the number of total iterations wouldn't exceed `nbIts`? Because if it does, then there is a bug somewhere. I think it is OK to panic in that case (it indicates error in HalfGCD algorithm), because otherwise we will get into a deadloop and ddos a service (which is worse than panic imo).

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+// GenNumber generates a random integer
+func GenNumber() gopter.Gen {
+	return func(genParams *gopter.GenParameters) *gopter.GenResult {
+		var prime, _ = new(big.Int).SetString("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffed", 16) // 2^255 - 19
+		elmt, _ := rand.Int(rand.Reader, prime)
```
ivokub, line 249: This way the results are not reproducible. Should use the `Rng` from genParams. See `gopter.Gen` documentation:
> IMPORTANT: Even though a generator is supposed to generate random values, it should do this in a reproducible way. Therefore a generator has to create the same result for the same GenParameters, i.e. ensure that you just use the RNG provided by GenParameters and no external one. If you just plug generators together you do not have to worry about this.

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+	var prime, _ = new(big.Int).SetString("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffed", 16) // 2^255 - 19
+	a0, _ := rand.Int(rand.Reader, prime)
+	a1, _ := rand.Int(rand.Reader, prime)
+	c0, _ := rand.Int(rand.Reader, prime)
+	c1, _ := rand.Int(rand.Reader, prime)
+	a := ComplexNumber{A0: *a0, A1: *a1}
+	c := ComplexNumber{A0: *c0, A1: *c1}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		HalfGCD(&a, &c)
```
ivokub, line 276: Afaik there was a recommendation to store the result in some variable to ensure that Go compiler doesn't optimize the call out? [See here](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go#:~:text=A%20note%20on%20compiler%20optimisations)

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+			d.Mul(a, b).Mul(&d, c)
+			e.Mul(c, b).Mul(&e, a)
+			return e.Equal(&d)
+		},
+		genE,
+		genE,
+		genE,
+	))
+
+	properties.Property("norm should always be positive", prop.ForAll(
```
ivokub, line 205: Norm can be zero for `0 + 0*\omega`? In that case imo the comparison below should be `a.Norm().Sign() >= 0`

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+			b.Add(a, zero)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("mul by one should leave element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			one := &ComplexNumber{
```
ivokub, line 149: I'd recommend to use
```
var one ComplexNumber
one.SetOne
b.Mul(a, one)
```

then we also check that we have `SetOne` correctly defined

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+			b.Conjugate(a).Conjugate(&b)
+			return a.Equal(&b)
+		},
+		genE,
+	))
+
+	properties.Property("add zero should leave element invariant", prop.ForAll(
+		func(a *ComplexNumber) bool {
+			var b ComplexNumber
+			zero := new(ComplexNumber)
```
ivokub, line 139: I'd also add test where the argument is obtained by `ComplexNumber.SetZero()` to ensure the method is correct.

field/eisenstein/eisenstein_test.go
```diff
@@ -0,0 +1,278 @@
...
+		},
+		genE,
+		genE,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+// GenNumber generates a random integer
+func GenNumber() gopter.Gen {
```
ivokub, line 246: Maybe we can provide the bound for the integer sampling as an argument? Or maybe just the number of bits for the integers and then sample from below `2^bound`?

ivokub (changes requested):
In general looks quite good already, I only added a few comments for general improvement, but nothing significant imo. See which ones make sense, you don't have to incorporate all.

ivokub (commented):
One thing though - I'm not sure if `gnark-crypto/field/eisenstein` is good place for it. Previously the GLV scalar splitting functions are defined in `gnark-crypto/ecc`, but I wouldn't want to overload the `ecc` package.

Maybe `gnark-crypto/utils` instead?

gbotrel:
>Maybe gnark-crypto/utils instead?
Not a big fan of "utils" package, I prefer option 1 on my side it's more self describing when a caller uses it

ivokub:
> > Maybe gnark-crypto/utils instead?
> 
> Not a big fan of "utils" package, I prefer option 1 on my side it's more self describing when a caller uses it

I completely agree, usually `utils` becomes something people throw stuff in and ... never use again :)

field/eisenstein/eisenstein.go
```diff
@@ -0,0 +1,155 @@
...
+func (z *ComplexNumber) Neg(x *ComplexNumber) *ComplexNumber {
+	z.A0.Neg(&x.A0)
+	z.A1.Neg(&x.A1)
+	return z
+}
+
+// Conjugate sets z to the conjugate of x, and returns z.
+func (z *ComplexNumber) Conjugate(x *ComplexNumber) *ComplexNumber {
+	z.A0.Sub(&x.A0, &x.A1)
+	z.A1.Neg(&x.A1)
```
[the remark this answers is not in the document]
yelhousni: the conjugate of `ω` is `ω^2=(-1-w)`
https://en.wikipedia.org/wiki/Eisenstein_integer

field/eisenstein/eisenstein.go
yelhousni: Good point should panic when `y==0` as in https://pkg.go.dev/math/big#Int.Quo

field/eisenstein/eisenstein.go
yelhousni: Something similar to this https://math.stackexchange.com/questions/600424/prove-that-the-gaussian-integers-ring-is-a-euclidean-domain but for Eisenstein (`j^2+j+1=0`)

field/eisenstein/eisenstein.go
yelhousni: Indeed

field/eisenstein/eisenstein.go
yelhousni: I'm currently revisiting the `HalfGCD` the condition should be related to the norm value and its size

field/eisenstein/eisenstein.go
```diff
@@ -114,31 +110,59 @@ func (z *ComplexNumber) Norm() *big.Int {
...
+	z.A0.Div(&z.A0, norm)
+	z.A1.Div(&z.A1, norm)
+	var t ComplexNumber
+	r.Sub(x, t.Mul(y, z))
+	return z, r
+}
+
+// Min returns the minimum of z... with respect to the norm.
+func Min(z ...*ComplexNumber) *ComplexNumber {
+	min := z[0]
```
ivokub, line 140: If `len(z) == 0` then get out of bounds panic. Dunno what is the best approach, return nil? Or explicit panic that the number of arguments to `Min` is zero. At some point we end up with a panic anyway, but otherwise it may be difficult to figure out what was the actual reason.

field/eisenstein/eisenstein_test.go
```diff
@@ -273,6 +272,6 @@ func BenchmarkHalfGCD(b *testing.B) {
 	c := ComplexNumber{A0: *c0, A1: *c1}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		HalfGCD(&a, &c)
+		_ = HalfGCD(&a, &c)
```
ivokub, line 275: I'm not sure it is sufficient. I think it is better to create a file level variable `var benchRet [3]*ComplexNumber` and then assign to it

From Dave Cheney's blog:
```go
var result int

func BenchmarkFibComplete(b *testing.B) {
        var r int
        for n := 0; n < b.N; n++ {
                // always record the result of Fib to prevent
                // the compiler eliminating the function call.
                r = Fib(10)
        }
        // always store the result to a package level variable
        // so the compiler cannot eliminate the Benchmark itself.
        result = r
}
```

ivokub (approved):
Now looks a lot better, thanks! I'm approving for now even though there are two more remarks. I'm not really sure the benchmark thing is an issue (but I guess it really depends on the optimization levels etc.) and the other input length check only provides better panic message.

ivokub:
Suggested edit:
```diff
diff --git a/field/eisenstein/eisenstein.go b/field/eisenstein/eisenstein.go
index 54bfde8af..2bf0dc46b 100644
--- a/field/eisenstein/eisenstein.go
+++ b/field/eisenstein/eisenstein.go
@@ -6,7 +6,16 @@ import (
 
 // A ComplexNumber represents an arbitrary-precision Eisenstein integer.
 type ComplexNumber struct {
-	A0, A1 big.Int
+	A0, A1 *big.Int
+}
+
+func (z *ComplexNumber) init() {
+	if z.A0 == nil {
+		z.A0 = new(big.Int)
+	}
+	if z.A1 == nil {
+		z.A1 = new(big.Int)
+	}
 }
 
 // String implements Stringer interface for fancy printing
@@ -16,55 +25,60 @@ func (z *ComplexNumber) String() string {
 
 // Equal returns true if z equals x, false otherwise
 func (z *ComplexNumber) Equal(x *ComplexNumber) bool {
-	return z.A0.Cmp(&x.A0) == 0 && z.A1.Cmp(&x.A1) == 0
+	return z.A0.Cmp(x.A0) == 0 && z.A1.Cmp(x.A1) == 0
 }
 
 // Set sets z to x, and returns z.
 func (z *ComplexNumber) Set(x *ComplexNumber) *ComplexNumber {
-	z.A0.Set(&x.A0)
-	z.A1.Set(&x.A1)
+	z.init()
+	z.A0.Set(x.A0)
+	z.A1.Set(x.A1)
 	return z
 }
 
 // SetZero sets z to 0, and returns z.
 func (z *ComplexNumber) SetZero() *ComplexNumber {
-	z.A0 = *big.NewInt(0)
-	z.A1 = *big.NewInt(0)
+	z.A0 = big.NewInt(0)
+	z.A1 = big.NewInt(0)
 	return z
 }
 
 // SetOne sets z to 1, and returns z.
 func (z *ComplexNumber) SetOne() *ComplexNumber {
-	z.A0 = *big.NewInt(1)
-	z.A1 = *big.NewInt(0)
+	z.A0 = big.NewInt(1)
+	z.A1 = big.NewInt(0)
 	return z
 }
 
 // Neg sets z to the negative of x, and returns z.
 func (z *ComplexNumber) Neg(x *ComplexNumber) *ComplexNumber {
-	z.A0.Neg(&x.A0)
-	z.A1.Neg(&x.A1)
+	z.init()
+	z.A0.Neg(x.A0)
+	z.A1.Neg(x.A1)
 	return z
 }
 
 // Conjugate sets z to the conjugate of x, and returns z.
 func (z *ComplexNumber) Conjugate(x *ComplexNumber) *ComplexNumber {
-	z.A0.Sub(&x.A0, &x.A1)
-	z.A1.Neg(&x.A1)
+	z.init()
+	z.A0.Sub(x.A0, x.A1)
+	z.A1.Neg(x.A1)
 	return z
 }
 
 // Add sets z to the sum of x and y, and returns z.
 func (z *ComplexNumber) Add(x, y *ComplexNumber) *ComplexNumber {
-	z.A0.Add(&x.A0, &y.A0)
-	z.A1.Add(&x.A1, &y.A1)
+	z.init()
+	z.A0.Add(x.A0, y.A0)
+	z.A1.Add(x.A1, y.A1)
 	return z
 }
 
 // Sub sets z to the difference of x and y, and returns z.
 func (z *ComplexNumber) Sub(x, y *ComplexNumber) *ComplexNumber {
-	z.A0.Sub(&x.A0, &y.A0)
-	z.A1.Sub(&x.A1, &y.A1)
+	z.init()
+	z.A0.Sub(x.A0, y.A0)
+	z.A1.Sub(x.A1, y.A1)
 	return z
 }
 
@@ -74,13 +88,14 @@ func (z *ComplexNumber) Sub(x, y *ComplexNumber) *ComplexNumber {
 //
 //	(x0+x1ω)(y0+y1ω) = (x0y0-x1y1) + (x0y1+x1y0-x1y1)ω
 func (z *ComplexNumber) Mul(x, y *ComplexNumber) *ComplexNumber {
+	z.init()
 	var t [3]big.Int
 	var z0, z1 big.Int
-	t[0].Mul(&x.A0, &y.A0)
-	t[1].Mul(&x.A1, &y.A1)
+	t[0].Mul(x.A0, y.A0)
+	t[1].Mul(x.A1, y.A1)
 	z0.Sub(&t[0], &t[1])
-	t[0].Mul(&x.A0, &y.A1)
-	t[2].Mul(&x.A1, &y.A0)
+	t[0].Mul(x.A0, y.A1)
+	t[2].Mul(x.A1, y.A0)
 	t[0].Add(&t[0], &t[2])
 	z1.Sub(&t[0], &t[1])
 	z.A0.Set(&z0)
@@ -97,12 +112,12 @@ func (z *ComplexNumber) Norm() *big.Int {
 	norm := new(big.Int)
 	temp := new(big.Int)
 	norm.Add(
-		norm.Mul(&z.A0, &z.A0),
-		temp.Mul(&z.A1, &z.A1),
+		norm.Mul(z.A0, z.A0),
+		temp.Mul(z.A1, z.A1),
 	)
 	norm.Sub(
 		norm,
-		temp.Mul(&z.A0, &z.A1),
+		temp.Mul(z.A0, z.A1),
 	)
 	return norm
 }
@@ -115,8 +130,8 @@ func (z *ComplexNumber) Quo(x, y *ComplexNumber) *ComplexNumber {
 	}
 	z.Conjugate(y)
 	z.Mul(x, z)
-	z.A0.Div(&z.A0, norm)
-	z.A1.Div(&z.A1, norm)
+	z.A0.Div(z.A0, norm)
+	z.A1.Div(z.A1, norm)
 	return z
 }
 
@@ -128,8 +143,8 @@ func (z *ComplexNumber) QuoRem(x, y, r *ComplexNumber) (*ComplexNumber, *Complex
 	}
 	z.Conjugate(y)
 	z.Mul(x, z)
-	z.A0.Div(&z.A0, norm)
-	z.A1.Div(&z.A1, norm)
+	z.A0.Div(z.A0, norm)
+	z.A1.Div(z.A1, norm)
 	var t ComplexNumber
 	r.Sub(x, t.Mul(y, z))
 	return z, r
diff --git a/field/eisenstein/eisenstein_test.go b/field/eisenstein/eisenstein_test.go
index 224497201..67624d3fe 100644
--- a/field/eisenstein/eisenstein_test.go
+++ b/field/eisenstein/eisenstein_test.go
@@ -246,7 +246,7 @@ func GenNumber(boundSize int64) gopter.Gen {
 		var bound big.Int
 		bound.Exp(big.NewInt(2), big.NewInt(boundSize), nil)
 		elmt, _ := rand.Int(genParams.Rng, &bound)
-		genResult := gopter.NewGenResult(*elmt, gopter.NoShrinker)
+		genResult := gopter.NewGenResult(elmt, gopter.NoShrinker)
 		return genResult
 	}
 }
@@ -257,7 +257,7 @@ func GenComplexNumber(boundSize int64) gopter.Gen {
 		GenNumber(boundSize),
 		GenNumber(boundSize),
 	).Map(func(values []interface{}) *ComplexNumber {
-		return &ComplexNumber{A0: values[0].(big.Int), A1: values[1].(big.Int)}
+		return &ComplexNumber{A0: values[0].(*big.Int), A1: values[1].(*big.Int)}
 	})
 }
 
@@ -268,8 +268,8 @@ func BenchmarkHalfGCD(b *testing.B) {
 	a1, _ := rand.Int(rand.Reader, prime)
 	c0, _ := rand.Int(rand.Reader, prime)
 	c1, _ := rand.Int(rand.Reader, prime)
-	a := ComplexNumber{A0: *a0, A1: *a1}
-	c := ComplexNumber{A0: *c0, A1: *c1}
+	a := ComplexNumber{A0: a0, A1: a1}
+	c := ComplexNumber{A0: c0, A1: c1}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		_ = HalfGCD(&a, &c)

```

field/eisenstein/eisenstein.go
yelhousni: Removed this method as it is no longer used

ivokub (approved):
Looks perfect now! Thanks for the fixes.

PR #544, opened 2024-09-22
feat: add vector.Sum using AVX512

gbotrel:
# Description

edit: this PR also includes #545 

Adds `vector.Sum(...)` using AVX512 instructions when available. 
Rough bench on `r7i.xlarge` instance gives 70-80% perf improvement for a vector with 1<<20 elements.

```
benchmark                        old ns/op     new ns/op     delta
BenchmarkElementVecOps/Sum-4     1689581       340038        -79.87%
BenchmarkElementVecOps/Sum-4     1679356       340570        -79.72%
```

Inspired with some modifications from https://github.com/a16z/vectorized-fields . 

TODO

- [ ] ensure https://github.com/a16z/vectorized-fields sets a License file
- [ ] ensure the Barrett reduction bounds are met (i.e. 2 conditional sub enough) with proof
- [ ] add test cases for edge cases in Barrett reduction


gbotrel:
moving back to draft, will consolidate with other vecops (refactoring + better tests + cleaner)

PR #547, opened 2024-09-28
perf: adds AVX512 implementations of vector.Sum, vector.InnerProduct + assembly refactor

gbotrel:
# Description

In this PR:
* factorization of existing assembly code (see #545 )
* adds `vector.Sum`, `vector.InnerProduct` , `vector.ScalarMul` and `vector.Mul` partially derived from Dag Arne Osvik's work in github.com/a16z/vectorized-fields
* tests

# Benchmarks

Benchmarks on size from 16 to 16M values. 

Seems results are better on AMD EPYC 9R14 (hpc7a, c7a, r7a) than on the intel xeon 8488c (r7i, ...). 

`hpc7a`
image

`r7i`
Screenshot 2024-10-03 at 2 29 02 PM
`c7a`
Screenshot 2024-10-03 at 2 28 49 PM


field/generator/generator.go
```diff
@@ -70,9 +70,12 @@ func GenerateFF(F *config.FieldConfig, outputDir string) error {
 
 	for _, of := range oldFiles {
 		_ = os.Remove(filepath.Join(outputDir, eName+of))
+		_ = os.Remove(filepath.Join(outputDir, of))
```
ivokub, line 73: Why this remove? For me it seems there shouldn't be any files named `_xxx.go`?

field/generator/asm/amd64/asm_macros.go
```diff
@@ -102,6 +117,20 @@ func (f *FFAmd64) GenerateDefines() {
 	f.WriteLn(buf.String())
 }
 
+func (f *FFAmd64) GenerateDefinesDeprecated(F *config.FieldConfig) {
```
ivokub, line 120: This is indicated as deprecated, but still used for some e2. Should be removed?

field/generator/generator_test.go
```diff
@@ -73,32 +78,58 @@ func TestIntegration(t *testing.T) {
...
+
+	for _, subDir := range subDirs {
+		// run go test in parallel
+		errGroup.Go(func() error {
+			cmd := exec.Command("go", "test")
+			cmd.Dir = subDir
+			var stdouterr strings.Builder
+			cmd.Stdout = &stdouterr
+			cmd.Stderr = &stdouterr
+			if err := cmd.Run(); err != nil {
```
ivokub, line 124: This approach doesn't kill other `go test` commands. Instead, we could create the new errGroup with `errgroup.WithContext` and provide the context to `cmd := exec.CommandContext`. The command should be killed when the errgroup context is cancelled (which is done when any `errGroup.Go` calls returns a non-nil error.

Or we could leave it as is - the test returns anywait later and it means the child processes should be killed afaik.

ivokub (approved):
Without trying to understand the assembly definitions, looks good to me. There are a few comments, but rather minor.

PR #564, opened 2024-12-04
feat: hash registry with statestorer

ivokub:
Continuation of https://github.com/Consensys/gnark-crypto/pull/560.


AlexandreBelling (approved):

internal/generator/crypto/hash/mimc/template/tests/mimc_test.go.tmpl
```diff
@@ -0,0 +1,114 @@
...
+	assert.NoError(t, err)
+	_, err = fs.ComputeChallenge("c0")
+	assert.NoError(t, err)
+}
+
+func TestByteOrder(t *testing.T) {
+	assert := require.New(t)
+
+	var buf [fr.Bytes]byte
+	// if the 31 first bytes are FF, it's a valid FF in little endian, but not in big endian
```
gbotrel, line 26: I guess here "31" should be {{ sub .Fr.Bytes 1 }} or something like that (unrelated to this PR I know :) )

internal/generator/crypto/hash/mimc/template/tests/mimc_test.go.tmpl
```diff
@@ -0,0 +1,114 @@
...
+
+	{
+		// hashing buf with little endian should succeed
+		mimcHash := mimc.NewMiMC(mimc.WithByteOrder(fr.LittleEndian))
+		_, err := mimcHash.Write(buf[:])
+		assert.NoError(err)
+	}
+
+	buf = [fr.Bytes]byte{}
+	// if the 31 bytes are FF, it's a valid FF in big endian, but not in little endian
```
gbotrel, line 50: same

hash/doc.go
```diff
@@ -1,5 +1,15 @@
...
 //
+// # Importing hash functions
+//
+// The package follows registration pattern for importing hash functions. To
+// import all known hash functions in gnark-crypto, import the
+// [github.com/consensys/gnark-crypto/hash/all] package in your code. To import
+// only a specific hash, then import the corresponding package directly, e.g.
+// [github.com/consensys/gnark-crypto/ecc/bn254/fr/mimc]. The import format should be:
+//
+//	import _ "github.com/consensys/gnark-crypto/ecc/bn254/fr/mimc"
```
gbotrel, line 11: so, package doc thing, do I understand correctly that the recommended way is to use directly the New function in the "curve-typed" import path and NOT use hash/ anymore ?

But that we keep hash/ for retro compat reasons? 

Can we make that explicit in the doc?

hash/hashes.go
```diff
@@ -57,26 +62,16 @@ var digestSize = []uint8{
...
-		return bw633.NewMiMC()
-	default:
-		panic("Unknown mimc ID")
+	if m < maxHash {
+		f := hashes[m]
+		if f != nil {
+			return f()
+		}
 	}
+	panic("requested hash function #" + m.String() + " not registered")
```
gbotrel, line 74: since it's going to break existing users code bases, can we make their life easier by augmenting this message with a suggestion containing the full "consider adding "import _ ..../strings.ToLower(m.String())/ ... " or whatever ?

gbotrel (changes requested):

hash/hashes.go
ivokub: Fixed

ivokub (commented):
Resolved issues. Also rebased on top of master due to bavard upgrade.

gbotrel (approved):

PR #585, opened 2024-12-18
Feat/write point slice ptr

Tabaie:
`Encoder.Write` now also accepts `*[]curve.G1Affine` and `*[]curve.G2Affine`.

It enables convenient functions like `refsSlice` seen here
https://github.com/Consensys/gnark/blob/chore/groth16-mpc/backend/groth16/bn254/mpcsetup/marshal.go#L60
, where the same piece of code orders a struct's fields for both serialization and deserialization.

### DIFF
--- a/internal/generator/ecc/template/marshal.go.tmpl
+++ b/internal/generator/ecc/template/marshal.go.tmpl
@@ -688,7 +688,7 @@ func (enc *Encoder) encode{{- $.Raw}}(v interface{}) (err error) {
 		}
 		return
 	case []G1Affine:
-		return enc.writeG1Slice{{- $.Raw}}(t)
+		return enc.writeG1Slice{{- $.Raw}}(t
 	case *[]G1Affine:
 		return enc.writeG1Slice{{- $.Raw}}(*t)
 	case []G2Affine:
--- a/internal/generator/ecc/template/marshal.go.tmpl
+++ b/internal/generator/ecc/template/marshal.go.tmpl
@@ -555,6 +555,48 @@ func (enc *Encoder) writeUint32(a uint32) error {
 	return err
 }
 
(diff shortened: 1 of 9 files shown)


internal/generator/ecc/template/marshal.go.tmpl
```diff
@@ -646,43 +688,13 @@ func (enc *Encoder) encode{{- $.Raw}}(v interface{}) (err error) {
...
 		return
 	case []G1Affine:
-		// write slice length
-		err = binary.Write(enc.w, binary.BigEndian, uint32(len(t)))
-		if err != nil {
-			return
-		}
-		enc.n += 4
-
-		var buf [SizeOfG1Affine{{- if $.Raw}}Uncompressed{{- else}}Compressed{{- end}}]byte
```
gbotrel, line 656: this PR loses the distinction of encoding compress vs non compressed points

gbotrel (changes requested):

internal/generator/ecc/template/marshal.go.tmpl
```diff
@@ -557,6 +557,48 @@ func (enc *Encoder) writeUint32(a uint32) error {
 
 {{ define "encode"}}
 
+func (enc *Encoder) writeG1Slice{{$.Raw}}(t []G1Affine) error {
```
gbotrel, line 560: {{$.Raw}}

What is that in the func name ? :) 

another way to do this PR & ensure it's the same code is just in the type switch, case *[]X: encode(*v)

but yeah as long as there is no regression OK on my side 👍

Tabaie: > another way to do this PR & ensure it's the same code is just in the type switch, case *[]X: encode(*v)
That's a good economic way to do it. I'll revert and do that.

gbotrel (approved):

PR #587, opened 2024-12-19
feat: cleaned iop

ThomasPiellard:
# Description

iop.Polynomial:
- removed blinding -> easier to deal with blinding manually with KZG(blinded(p)) = Kzg(p) + KZG(blinding_factor)
- code no longer panics when polynomials are evaluated LAGRANGE or LAGRANGE_COSET basis

(restoring an old branch and PR, with the updated code)

## Type of change

# How has this been tested?

- [x] Test TestEvaluation

# Checklist:

### DIFF
--- a/internal/generator/iop/template/polynomial.go.tmpl
+++ b/internal/generator/iop/template/polynomial.go.tmpl
@@ -54,9 +54,9 @@ var (
 // default blindedSize=Size, until the polynomial is blinded.
 type Polynomial struct {
 	*polynomial
-	shift       int
-	size        int
-	blindedSize int
+	shift int
+	size  int
+	coset fr.Element // needed for evaluating the polynomial when it is expressed in Lagrange shifted basis
 }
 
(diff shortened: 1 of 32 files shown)


internal/generator/iop/template/polynomial.go.tmpl
```diff
@@ -101,69 +94,31 @@ func (p *Polynomial) SetSize(size int) {
...
-	p.blindedSize = newSize
-
-	return p
-}
-
 // Evaluate evaluates p at x.
 // The code panics if the function is not in canonical form.
 func (p *Polynomial) Evaluate(x fr.Element) fr.Element {
 
+	if p.Basis == LagrangeCoset {
```
gbotrel, line 101: question; seems p.Basis is only used in the iop package. Do we need it exposed? 

Wondering if this should be private to avoid accidental changes in the basis by copies or whatever, that could lead to a state where p.Basis is wrongly set and the x.Div(x, p.Coset) is a div by 0 that fails silently.

PR #594, opened 2025-01-03
chore: docs fix spelling issues 

sky-coderay:


ivokub:
Thanks for the contribution. However we have decided against accepting contributions from users without public activity.

PR #607, opened 2025-01-15
build: improve go generate generation time by checking git status

gbotrel:
# Description

Not fail proof, but added a `git.HasChanges(..)` internal api, which check if templates have changes before regen files.

The CI has a special build tag that bypasses this, in case some templates end up having some weird dependency chain, we should detect it.

Speeds up dev process when changing only one part of the code base, doesn't regenerate everything.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,7 +19,7 @@ jobs:
       run: if [[ -n $(gofmt -l .) ]]; then echo "please run gofmt"; exit 1; fi
     - name: generated files should not be modified
       run: |
-        go generate ./...
+        go generate -tags=forcegen ./...
         git update-index --assume-unchanged go.mod
         git update-index --assume-unchanged go.sum
         if [[ -n $(git status --porcelain) ]]; then echo "git repo is dirty after running go generate -- please don't modify generated files"; echo $(git diff);echo $(git status --porcelain); exit 1; fi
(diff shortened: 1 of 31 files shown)


.github/workflows/pr.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       run: if [[ -n $(gofmt -l .) ]]; then echo "please run gofmt"; exit 1; fi
     - name: generated files should not be modified
       run: |
-        go generate ./...
+        go generate -tags=forcegen ./...
```
ivokub, line 22: I'm not sure we can use `-tags=forcegen` that easily. This imo defines the build tag for building the generator, but not how we run the commands to actually run later. At least when I'm running locally `go generate -tags=forcegen ./...` then it doesn't force the generation.

ivokub: Maybe we can get it working though - in the packages where we have directives
```
//go:generate go run main.go
```
we define a global constant which is defined when we call `go generate -tags=forcegen ./...` and then refer to it in the subgenerators?

ivokub (changes requested):
Looks very nice PR, however there are a few issues atm (more specific in comments):
- we don't handle errors when calling `git status`. It may lead to not generating when we need to in build environment.
- imo `forcegen` tag doesn't work right now as is. We still run `go run main.go` skipping force regeneration.
- currently this approach is a bit tricky in cases where we make a change to the template, generate and then revert the changes to template. Then it is really difficult to revert the changes in the generated files. Maybe we should add some message to the console or change the default behaviour? E.g. instead of `forcegen` tag we have `lazygen` etc.?
- one of the issues I have had is when I have changed the path of generates files. Then when generating the old files still stay around and I have to manually remove them. Maybe it would be worth it to keep track which files have been generated and remove them automatically somehow?

gbotrel:
indeed, I rushed that one, will think of something better and test more, closing for now :)

PR #608, opened 2025-01-16
feat(bw6-761): Miller loop using direct sextic extension

yelhousni:
# Description

This corresponds to the [gnark circuit](https://github.com/Consensys/gnark/blob/cfa30d90b1a438da4551c01c88ddea0c38aacb37/std/algebra/emulated/sw_bw6761/pairing.go#L253) and is needed to compute the same `residueWitness` in- and out-circuit for https://github.com/Consensys/gnark/pull/1365.

## Type of change

# How has this been tested?

Added tests for the new `E6D` struct in `internal/fptower/e6_direct_test.go` and pairing computation using towered and direct extensions in `pairing_test.go`.

`MillerLoopDirect` is much slower than `MillerLoop` but it is only used in the hint computation for the gnark circuit.

```
benchmark                 tower ns/op   direct ns/op     delta
BenchmarkMillerLoop-8     1195105       3922503       +228.21%
```

# Checklist:

### DIFF
--- a/ecc/bw6-761/pairing.go
+++ b/ecc/bw6-761/pairing.go
@@ -653,14 +653,18 @@ func (p *G2Affine) tangentCompute(evaluations *LineEvaluationAff) {
 // Alg.2 in https://eprint.iacr.org/2021/1359.pdf
 // Eq. (6') in https://hackmd.io/@gnark/BW6-761-changes
 func MillerLoopDirect(P []G1Affine, Q []G2Affine) (GT, error) {
+	// check input size match
+	n := len(P)
+	if n == 0 {
+		return GT{}, errors.New("invalid inputs sizes")
+	}
+
 	lines := make([][2][len(LoopCounter) - 1]LineEvaluationAff, 0, len(Q))
 	for _, qi := range Q {
 		lines = append(lines, PrecomputeLines(qi))
 	}
 
-	// check input size match
-	n := len(P)
-	if n == 0 || n != len(lines) {
+	if n != len(Q) {
 		return GT{}, errors.New("invalid inputs sizes")
 	}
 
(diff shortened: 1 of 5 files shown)


ecc/bw6-761/pairing.go
```diff
@@ -640,3 +640,97 @@ func (p *G2Affine) tangentCompute(evaluations *LineEvaluationAff) {
...
+// Alg.2 in https://eprint.iacr.org/2021/1359.pdf
+// Eq. (6') in https://hackmd.io/@gnark/BW6-761-changes
+func MillerLoopDirect(P []G1Affine, Q []G2Affine) (GT, error) {
+	lines := make([][2][len(LoopCounter) - 1]LineEvaluationAff, 0, len(Q))
+	for _, qi := range Q {
+		lines = append(lines, PrecomputeLines(qi))
+	}
+
+	// check input size match
+	n := len(P)
```
ivokub, line 662: Would move this sanity check part before the line precomputation. Then you can just measure `n != len(Q)`. Then we fail early before computing the lines.

ecc/bw6-761/pairing.go
```diff
@@ -640,3 +640,97 @@ func (p *G2Affine) tangentCompute(evaluations *LineEvaluationAff) {
...
+	for k := 0; k < n; k++ {
+		xNegOverY[k].Mul(&P[k].X, &yInv[k]).
+			Neg(&xNegOverY[k])
+	}
+
+	// f_{a0+λ*a1,Q}(P)
+	var result fptower.E6D
+	result.SetOne()
+
+	for i := 188; i >= 0; i-- {
```
ivokub, line 698: Maybe loop initial variable `i = len(LoopCounter)-2` instead?

ivokub (commented):
Small cosmetic comments, otherwise looks perfect. Didn't check fully the direct mulMontgomery equations, but as the tests check compatibility with fp2fp3 tower multiplication, then should be good.

Thanks!

ivokub (approved):

PR #615, opened 2025-01-22
Perf: Parallelize subgroup checks in KZG MPC setup

Tabaie:


internal/generator/kzg/template/mpcsetup.go.tmpl
```diff
@@ -113,12 +116,26 @@ func (s *MpcSetup) Verify(next *MpcSetup) error {
...
+	// TODO @Tabaie replace with batch subgroup check
+	n := len(next.srs.Pk.G1) - 1
+	fail := -1
+	wp := utils.NewWorkerPool()
+	defer wp.Stop()
+
+	wp.Submit(n, func(start, end int) {
+		for i := start; i < end; i++ {
+			if !next.srs.Pk.G1[i+1].IsInSubGroup() {
+				fail = i + 1
```
gbotrel, line 131: that does not work; fail can be accessed by multiple go routines .
Either use a chan of error that you check at the end, or at atomic counter that you increment

gbotrel (changes requested):

PR #623, opened 2025-02-03
feat: Poseidon2 Hash Instantiation for BLS12-377

Tabaie:
# Description

This PR provides a generic Merkle Damgard construction, an instantiation of Poseidon2 as a hash for BLS12-377, and GKR gates for the same curve.

Fixes # (issue)

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

Please describe the benchmarks that you ran to verify your changes.

- [ ] Benchmark A, on Macbook pro M1, 32GB RAM
- [ ] Benchmark B, on x86 Intel xxx, 16GB RAM

# Checklist:


ivokub (approved):
It seemed that we didn't have the diagonal matrices if `T != 2,3`, so I explicitly call panic for now to ensure this mode is not used. Previously we essentially multiplied by uninitialized (zero) matrix which is probably insecure.

I also refactored a bit so that we can better refer to the parameters and round constants from gnark side. I also made the seed fully deterministic as this is what the reference implementation did and imo leaves less room for making errors.

Finally, renamed some of the arguments/field names to better describe what they are (as previously we would have to check the actual implementation instead of the package documentation).

I also separated the GKR gates into a separate package so that we would only need to load it when needed. I also recommend using the registry-based approach as currently the `gkr.Gates` access is not synchronized which may lead to inconsistent state.

I see that you have already started another PR #628 which assumes this PR. Feel free to revert the commits but I think it could be useful for more generic usage.

ecc/bls12-377/fr/poseidon2/gkrgates/gkrgates.go
```diff
@@ -0,0 +1,193 @@
...
+
+			gateNameX := func(i int) string {
+				return fmt.Sprintf("x-round=%d%s", i, p.String())
+			}
+			gateNameY := func(i int) string {
+				return fmt.Sprintf("y-round=%d%s", i, p.String())
+			}
+
+			fullRound := func(i int) {
+				gkr.Gates[gateNameX(i)] = &extKeySBoxGate{
```
ivokub, line 144: I copy-pasted this part - but here you have unsynchronized access to `gkr.Gates`. It may lead to situations where several places the map is updated at the same time, leading to only the last write to succeed.

Better approach is to use the registry approach where `gkr` package provides function `RegisterGate` and performs assignment in synchronous manner.

Tabaie: Agreed. Will address it in the next PR.

PR #634, opened 2025-02-13
InterpolateOnRange refactor

Tabaie:
1. Move `InterpolateOnRange` from `multilin.go` to `polynomial.go`
2. Remove precomputation in package `init`
3. Synchronize cached values


internal/generator/polynomial/template/polynomial.go.tmpl
```diff
@@ -200,3 +201,120 @@ func (p Polynomial) Text(base int) string {
...
+		res.Add(res, temp)
+	}
+
+	return res
+}
+
+// lagrange bases used by InterpolateOnRange
+var (
+	lagrangeBasisLock sync.RWMutex
+	lagrangeBasis     [][]Polynomial
```
gbotrel, line 231: I would just use a sync.Map here. Unless we expect a scenario were lots of go routine wants to read while others may initialize entries with different domainSize, a sync.Map Load / Store is clean enough

gbotrel (changes requested):

internal/generator/polynomial/template/polynomial.go.tmpl
gbotrel: Is it called in a "very concurrent way" ? because w/ current design ,the cache if you have 2 go routines querying domain size 100, not init yet; they will both hit the Rlock, and both compute the entry, and both update the lagrangeBasis array. 

so either mutex everything (i;e. the full getter is protected by a Lock / Unlock), or have one lock per entry if we expect concurrent accesses  on different domain sizes

Tabaie: We expect very few writes (identical ones too, so imo "pre" computing size 100 twice is not a huge deal) and many, many reads that shouldn't block each other. I think a regular mutex wouldn't be good for the use case but sync.Map would be.

gbotrel (approved):

PR #658, opened 2025-03-07
test(bn254, bls12-381): test points intentionally not on sugroups G1/2

yelhousni:
# Description

- Generate test points that are intentionally on the G1/2 cofactor-torsions for fuzzing purposes. 
- Renamed some test methods to make more sense.

This is useful for geth fuzzer for bn254 and bls12-381. We could also generify the test for other curves but it's not a priority now IMO.

## Type of change

# How has this been tested?

Last test in `TestIsOnG2` generates points (using `fuzzCofactorOfG*Jac`) that pass the `IsOnCurve` test but not `IsInSubGroup` test.

N/A

# Checklist:

### DIFF
--- a/ecc/bls12-381/g1_test.go
+++ b/ecc/bls12-381/g1_test.go
@@ -872,10 +872,6 @@ func fuzzCofactorOfG1(f fp.Element) G1Jac {
 	return res
 }
 
-func GeneratePointNotInG1(f fp.Element) G1Jac {
-	return fuzzCofactorOfG1(f)
-}
-
 func fuzzG1Jac(p *G1Jac, f fp.Element) G1Jac {
 	var res G1Jac
 	res.X.Mul(&p.X, &f).Mul(&res.X, &f)
--- a/ecc/bls12-381/g1_test.go
+++ b/ecc/bls12-381/g1_test.go
@@ -107,7 +107,7 @@ func TestIsOnG1(t *testing.T) {
 	))
 	properties.Property("[BLS12-381] IsInSubGroup should return false for a point on the cofactor-torsion", prop.ForAll(
 		func(a fp.Element) bool {
-			op := fuzzCofactorOfG1Jac(a)
+			op := fuzzCofactorOfG1(a)
 			return op.IsOnCurve() && !op.IsInSubGroup()
 		},
 		GenFp(),
@@ -859,7 +859,7 @@ func BenchmarkG1AffineDouble(b *testing.B) {
 		a.Double(&a)
 	}
 }
-func fuzzCofactorOfG1Jac(f fp.Element) G1Jac {
+func fuzzCofactorOfG1(f fp.Element) G1Jac {
 	var res, jac G1Jac
 	aff := MapToCurve1(&f)
 	g1Isogeny(&aff)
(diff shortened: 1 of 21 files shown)


holiman:
This is very useful, thanks! 

Currently, as far as I can tell this is not exported? 

t would be highly useful for external callers to be able to use this, to create points that are not in the right subgroups. Ideally using a custom source, so that fuzzing can produce points reproducably. 

Either maybe something like `GeneratePointInWrontgSubgroup( io.Reader)` or something which converts a point into a "not-on-subgroup-point) somehow. 

Alternatively: not expose the full "generate a point not on subgroup" but the individual smaller parts, making it easier for the outer caller to construct it by themselves. 

But maybe I'm just misunderstanding, and this is already doable? I'm not a cryptographer, this stuff is hard for me :)

yelhousni:
> This is very useful, thanks!
> 
> Currently, as far as I can tell this is not exported?
> 
> t would be highly useful for external callers to be able to use this, to create points that are not in the right subgroups. Ideally using a custom source, so that fuzzing can produce points reproducably.
> 
> Either maybe something like `GeneratePointInWrontgSubgroup( io.Reader)` or something which converts a point into a "not-on-subgroup-point) somehow.
> 
> Alternatively: not expose the full "generate a point not on subgroup" but the individual smaller parts, making it easier for the outer caller to construct it by themselves.
> 
> But maybe I'm just misunderstanding, and this is already doable? I'm not a cryptographer, this stuff is hard for me :)

Hey!
Would something like:
```go
func GeneratePointInWrongSubgroup(r io.Reader) (*G1Jac, error) {
        var f fp.Element
        var p G1Jac
        _, err := f.SetRandom()
        if err != nil {
                return &p, err
        }
        p = fuzzCofactorOfG1Jac(f)
        return &p, nil
}
```
be sufficient?
where `SetRandom()` is derived from [go/src/crypto/rand](https://github.com/golang/go/blob/bc5f4a555e933e6861d12edba4c2d87ef6caf8e6/src/crypto/rand/util.go#L67).

We can export this method or similar if so.

holiman:
Iiuc, using SetRandom like that will make it difficult to emit reproducible points.

holiman:
How about taking an fp.Element as input?

ecc/bls12-381/g1_test.go
```diff
@@ -872,6 +872,10 @@ func fuzzCofactorOfG1Jac(f fp.Element) G1Jac {
 	return res
 }
 
+func GeneratePointNotInG1(f fp.Element) G1Jac {
```
holiman, line 875: These must not be declared in the `_test.go` files. Those are only compiled when testing the package they are part of, and not part of the externally accesible set

ecc/bls12-381/g2_test.go
```diff
@@ -841,6 +848,21 @@ func BenchmarkG2AffineDouble(b *testing.B) {
...
+	// ψ(p)-[x₀]P = [r]p
+	res.mulBySeed(&jac)
+	jac.psi(&jac)
+	res.AddAssign(&jac)
+	return res
+}
+
+func GeneratePointNotInG2(f fptower.E2) G2Jac {
+	return fuzzCofactorOfG2(f)
+}
```
holiman, line 865: The `fptower.E2` is internal, so this public accessor needs to look like e.g. 
```golang
func GeneratePointNotInG2(f1, f2 fp.Element) G2Jac {
	return fuzzCofactorOfG2(E2{f1, f2})
}
```

ecc/bls12-381/g1_test.go
yelhousni: true! my bad

gbotrel (approved):

PR #661, opened 2025-03-11
test: clean up templates and add interop for poseidon2 babybear against plonky3 impl

gbotrel:


yelhousni (approved):
looks good to me! it matches plonky3/horizen and code generation is cleaner :)
I ran the [sage script](https://github.com/HorizenLabs/poseidon2/blob/main/poseidon2_rust_params.sage) from Hoizen and and it output the same constants you have in this PR. So we can also cross check koala bear. Here is the corresponding rust file: https://gist.github.com/yelhousni/4ebb017ed55425357c188be3a11b5a64

PR #667, opened 2025-03-22
perf: remove useless copies in e4 mul by elem

gbotrel:
# Description

At least for small fields, the copy is useless / hurts perf. Likely same for other towers?

```
benchmark                      old ns/op     new ns/op     delta
BenchmarkE4MulByElement-16     6.74          5.95          -11.79%
BenchmarkE4MulByElement-16     6.76          5.86          -13.29%
```


yelhousni (approved):
lgtm!
probably the same for all `e2.go`'s e.g. https://github.com/Consensys/gnark-crypto/blob/c2b5c443738fffe014c3e5e5e9c237d2d460e8ba/ecc/bn254/internal/fptower/e2.go#L133

PR #670, opened 2025-03-27
Remove GKR from gnark-crypto

Tabaie:
As a proof system, it is better housed in gnark. Companion gnark PR: https://github.com/Consensys/gnark/pull/1467


gbotrel (approved):
so, I'm going to approve, but I think the `polynomial` package should only be defined in gnark-crypto and not "duplicated with different API" in gnark..

i.e. sumcheck / gkr in gnark ok, I think while we iterate on it it makes sense. But the low level stuff like polynomial should be stable enough, sturdy in gnark-crypto imo.

PR #674, opened 2025-04-07
refactor: hash to curve

ivokub:
# Description

This PR exposes the primitive methods of hash-to-curve so that it can be implemented in gnark. Also separated the code generation into separate method.

Kept the current public methods intact, so it is fully backwards compatible.

## Type of change

- [x] refactor

# Checklist:


ivokub:
Currently not yet ready, I'll confirm it would also be useful for map to G2 in gnark.

ivokub:
@ThomasPiellard, @Tabaie - ready for review now.

ecc/bls12-377/hash_to_g1.go
```diff
@@ -175,38 +87,38 @@ func MapToCurve1(u *fp.Element) G1Affine {
 	var tv3 fp.Element
 	var tv4 fp.Element
 	tv4.SetOne()
-	tv3.Add(&tv2, &tv4)                // 5.  tv3 = tv2 + 1
-	tv3.Mul(&tv3, &sswuIsoCurveCoeffB) // 6.  tv3 = B * tv3
+	tv3.Add(&tv2, &tv4)               // 5.  tv3 = tv2 + 1
+	tv3.Mul(&tv3, &g1sswuCurveBCoeff) // 6.  tv3 = B * tv3
```
Tabaie, line 91: iirc this is a coefficient of an isogenous curve, not our curve itself. That should be reflected in its name.

ecc/bls12-377/hash_to_g1_test.go
```diff
@@ -30,7 +32,7 @@ func TestG1SqrtRatio(t *testing.T) {
 		func(u fp.Element, v fp.Element) bool {
 
 			var seen fp.Element
-			qr := g1SqrtRatio(&seen, &u, &v) == 0
+			qr := hash_to_curve.G1SqrtRatio(&seen, &u, &v) == 0
```
Tabaie, line 35: This whole test should be in the new `hash_to_curve` package.

internal/generator/ecc/template/point.go.tmpl
```diff
@@ -22,6 +22,9 @@ import (
 	{{else}}
 	"github.com/consensys/gnark-crypto/ecc/{{.Name}}/fp"
 	{{- end}}
+	{{- if eq .Name "bls12-381"}}
+	"github.com/consensys/gnark-crypto/ecc/{{.Name}}/hash_to_curve"
```
Tabaie, line 26: Why only for bls12-381?

Tabaie (changes requested):
I don't think it makes a lot of sense to have `hash_to_curve` source files in both packages. imo every hash to curve function should be moved to the new package and only add a few wrappers to the root package, thus removing `hash_to_g{{i}}.go` from the root package. This way we can avoid making all those internal functions public as well.

internal/generator/ecc/template/point.go.tmpl
ivokub: We need hash to curve only for generating points not in the subgroup. It is implemented only for fuzzing purposes in geth. And as EVM supports only BN254 and BLS12-381, where BN254 is prime order, then it is implemented only for BLS12-381.

ivokub:
> I don't think it makes a lot of sense to have `hash_to_curve` source files in both packages. imo every hash to curve function should be moved to the new package and only add a few wrappers to the root package, thus removing `hash_to_g{{i}}.go` from the root package. This way we can avoid making all those internal functions public as well.

The goal of the PR is to make some internal things public so that we can reuse in gnark:
- isogeny map coefficients
- isogenous curve coefficients
- sqrtRatio method
- clear cofactor method (for tests)

For the `MapToG`, `MapToCurve`, `EncodeToG` and `HashToG`, I tried moving them to the `hash_to_curve` package, but it resulted in import cycle, as it depends on `G1Affine` and `G1Jac`, so root package would import `hash_to_curve` which would then import root package again.

For the Isogeny methods I overcame that by working directly with the point coordinates. I could have done it also for `MapToG`, `MapToCurve` and `EncodeToG`, but as `HashToG` explicitly depends on curve arithmetic, then I cannot use only the coordinates. And if I need to keep `HashToG` in the root package, then it also made sense to keep rest.

ecc/bls12-377/hash_to_g1_test.go
ivokub: I'll move this particular test. I could in theory also move rest of the tests (for map and hash methods), by defining the test file to be in `{{ curve }}_test` package, but it would break code coverage tools as they only test the coverage within the same package.

ivokub: Moved

ecc/bls12-377/hash_to_g1.go
ivokub: Renamed

Tabaie (approved):
Thanks looks good

PR #653, opened 2025-02-27
feat: Add options for custom hash function and randomness source in MPC setup ceremonies

VolodymyrBg:
### Description
This PR adds support for custom hash functions and randomness sources in MPC setup ceremonies, as requested in issue #626. These changes enable in-SNARK verification of setup ceremonies and make debugging easier by allowing users to provide their own implementations.
The implementation includes:
New interfaces for hash functions and randomness sources
Default implementations that maintain backward compatibility
Updated MPC setup templates to use these interfaces
Documentation and examples showing how to use custom implementations
Fixes #626
### Type of change
[x] New feature (non-breaking change which adds functionality)
[ ] Bug fix (non-breaking change which fixes an issue)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[ ] This change requires a documentation update
### How has this been tested?
[x] Added example test that demonstrates using custom hash functions and randomness sources
[x] Verified that default implementations maintain backward compatibility
[x] Manual testing with custom implementations
### How has this been benchmarked?
No benchmarks were necessary as this change adds optional functionality that doesn't affect performance when using default implementations.
### Checklist:
[x] I have performed a self-review of my code
[x] I have commented my code, particularly in hard-to-understand areas
[x] I have made corresponding changes to the documentation
[x] I have added tests that prove my fix is effective or that my feature works
[x] I did not modify files generated from templates
[x] golangci-lint does not output errors locally
[x] New and existing unit tests pass locally with my changes
[x] Any dependent changes have been merged and published in downstream modules

PR #678, opened 2025-04-23
feat: Options for custom hash function and randomness source in MPC setup ceremonies

### LINKED ISSUE #626: Options for custom hash function and randomness source in MPC setup ceremonies
Based on @ivokub's feedback on https://github.com/Consensys/gnark/pull/1372

A custom hash function (i.e. option to use MiMC, Poseidon and the like) can enable in-SNARK verification of setup ceremonies.
A custom randomness source can make debugging easier.

crStiv:
## Description

This PR adds support for custom hash functions and randomness sources in MPC setup ceremonies, as requested in issue #626. These changes enable in-SNARK verification of setup ceremonies and make debugging easier by allowing users to provide their own implementations.

The implementation includes:
- New interfaces for hash functions and randomness sources
- Thread-safe global options using the functional options pattern
- Default implementations that maintain backward compatibility
- Updated MPC setup templates to use these interfaces
- Added `SetRandomWithSource` method to `fr.Element`
- Documentation and examples showing how to use custom implementations

### Custom Hash Function

```go
// Create a custom hash function
customHashFunc := func(msg, dst []byte) (curve.G2Affine, error) {
    // Use SHA-256 as an example
    h := sha256.New()
    h.Write(msg)
    h.Write(dst)
    digest := h.Sum(nil)
    
    // Use the built-in HashToG2 with our custom digest
    return curve.HashToG2(digest, []byte{0x01})
}

// Configure MPC with the custom hash function
mpcsetup.ConfigureMPC(mpcsetup.WithHashToG2(customHashFunc))
```

### Custom Randomness Source

```go
// Create a deterministic random source for debugging
deterministicSeed := []byte("fixed seed for deterministic randomness")
customRandomReader := bytes.NewReader(deterministicSeed)

customRandomSource := func() (io.Reader, error) {
    customRandomReader.Reset(deterministicSeed)
    return customRandomReader, nil
}

// Configure MPC with the custom randomness source
mpcsetup.ConfigureMPC(mpcsetup.WithRandomSource(customRandomSource))
```

## Implementation Note

This implementation specifically addresses the feedback from the previous PR (#653):

1. Uses the functional options pattern as suggested in the feedback
2. Ensures thread-safety for global hash function and randomness variables with mutex locks
3. Maintains backward compatibility with existing code

## Type of change

[x] New feature (non-breaking change which adds functionality)  
[ ] Bug fix (non-breaking change which fixes an issue)  
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)  
[ ] This change requires a documentation update

## How has this been tested?

- Added comprehensive test examples that demonstrate using custom hash functions and randomness sources
- Verified that default implementations maintain backward compatibility
- Tested thread-safety with concurrent configurations

Fixes #626

### DIFF
--- a/internal/generator/mpcsetup/template/mpcsetup.go.tmpl
+++ b/internal/generator/mpcsetup/template/mpcsetup.go.tmpl
@@ -14,12 +14,12 @@ import (
 // Generate R∈𝔾₂ as Hash(gˢ, challenge, dst)
 // it is to be used as a challenge for generating a proof of knowledge to x
 // π ≔ x.R; e([1]₁, π) =﹖ e([x]₁, R)
-func pokBase(xG curve.G1Affine, challenge []byte, dst byte) curve.G2Affine {
+func (ceremony *Ceremony) pokBase(xG curve.G1Affine, challenge []byte, dst byte) curve.G2Affine {
 	var buf bytes.Buffer
 	buf.Grow(len(challenge) + curve.SizeOfG1AffineUncompressed)
 	buf.Write(xG.Marshal())
 	buf.Write(challenge)
-	xpG2, err := mpcHash(buf.Bytes(), []byte{dst})
+	xpG2, err := ceremony.Hash(buf.Bytes(), []byte{dst})
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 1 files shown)


internal/generator/field/template/element/custom_random.go.tmpl
```diff
@@ -0,0 +1,32 @@
+// SetRandomWithSource sets z to a uniform random value in [0, q) using the provided random source.
+//
+// This might error only if reading from source errors,
+// in which case, value of z is undefined.
+func (z *{{.ElementName}}) SetRandomWithSource(source io.Reader, l, k int, b uint) (*{{.ElementName}}, error) {
+	// l is total number of bytes needed to reconstruct z
+	// k is the maximum byte length needed to encode a value < q
+	// b is the number of bits in the most significant byte of q-1
+
+	var bytes [{{mul 8 .NbWords}}]byte
```
Tabaie, line 10: Better to use `fieldPackage.Bytes`  instead of the multiplication here.

Tabaie:
Thank you for the contribution! Am I understanding correctly that the hash and rand settings are global? Since the `mpcsetup` package is so new, I wouldn't worry about backwards compatibility and change the interface in a way that  plays naturally with local settings.

Also, please make sure to run `go generate` and commit the generated files.

Tabaie (changes requested):

PR #679, opened 2025-04-26
feat: add vector operation on koalabear extension

gbotrel:
# Description

WIP, no tests, no AVX acceleration, just to iterate on the interface and actual need for acceleration.


field/generator/internal/templates/extensions/e4.vector.go.tmpl
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+func (vector *Vector) AddMixed(a Vector, b fr.Vector) {
+	if len(a) != len(b) || len(a) != len(*vector) {
+		panic("vector.AddMixed: vectors don't have the same length")
+	}
+	for i := range a {
+		(*vector)[i].Set(&a[i])
+		(*vector)[i].B0.A0.Add(&(*vector)[i].B0.A0, &b[i])
```
AlexandreBelling, line 25: Does that mean that we hard-cast vectors of base-field elements into vectors of extension field elements?

PR #686, opened 2025-05-16
feat(glv): split scalar strict bounds

yelhousni:
# Description

We re-introduce the original `SplitScalar()` removed in #213 (now renamed `SplitScalarStrict()`). For gnark-crypto we continue to use the current `SplitScalar()` which replaces divisions by right-shifts but we need `SplitScalarStrict()` for gnark circuits.

## Type of change

# How has this been tested?

`TestSplitting` in `utils_test.go` tests both `SplitScalar()` and `SplitScalarStrict()`.

```
goos: darwin
goarch: arm64
pkg: github.com/consensys/gnark-crypto/ecc
cpu: Apple M1
BenchmarkSplitting256
BenchmarkSplitting256-8                  4004929               253.8 ns/op
BenchmarkSplittingStrict256
BenchmarkSplittingStrict256-8            1352343               884.2 ns/op
```

# Checklist:

### DIFF
--- a/ecc/utils_test.go
+++ b/ecc/utils_test.go
@@ -78,7 +78,7 @@ func TestNafDecomposition(t *testing.T) {
 func TestSplitting(t *testing.T) {
 	t.Parallel()
 
-	var lambda, r, s, _s, zero big.Int
+	var lambda, r, s, _s1, _s2, zero big.Int
 	var l Lattice
 
 	r.SetString("21888242871839275222246405745257275088548364400416034343698204186575808495617", 10)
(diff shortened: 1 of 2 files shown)


feltroidprime:
Nice, does this mean the 4 scalars will be `i64`  in the glv+fakeglv method ?

yelhousni:
> Nice, does this mean the 4 scalars will be `i64` in the glv+fakeglv method instead of ~`i73` ?

unfortunately not. but we can probably save 1 bit.

ecc/utils_test.go
```diff
@@ -78,7 +78,7 @@ func TestNafDecomposition(t *testing.T) {
 func TestSplitting(t *testing.T) {
```
ivokub, line 78: Could we maybe run the test in gopter? Imo having a test with hardcoded values isn't that useful?

PR #690, opened 2025-06-02
[secp256k1]: replace outdated link to article

gap-editor:


ivokub:
> Change looks good to me and the current link is a dead end. @yelhousni can you confirm the paper is right?

Found it in the web archive - https://web.archive.org/web/20240418115219/https://www.cosic.esat.kuleuven.be/nessie/reports/phase2/GLV.pdf. Indeed it is the same paper. Thanks for the contribution!

PR #699, opened 2025-06-23
fix: MerkleDamgardHasher.Sum

Tabaie:
Updates the Sum function to match expected interface behavior. Adds a test to ascertain this was done correctly.

### DIFF
--- a/hash/interface.go
+++ b/hash/interface.go
@@ -18,6 +18,8 @@ type StateStorer interface {
 // Compressor is a 2-1 one-way function. It takes two inputs and compresses
 // them into one output. The inputs and outputs are all of the same size, which
 // is the block size. See [BlockSize].
+// The function must be stateless. Any two calls with the same input must yield the same output.
+// It must furthermore not modify its input, even temporarily.
 //
 // NB! This is lossy compression, meaning that the output is not guaranteed to
 // be unique for different inputs. The output is guaranteed to be the same for
--- a/hash/hash_test.go
+++ b/hash/hash_test.go
@@ -1,0 +1,41 @@
+package hash_test
+
+import (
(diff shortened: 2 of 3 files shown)


hash/interface.go
```diff
@@ -18,6 +18,8 @@ type StateStorer interface {
 // Compressor is a 2-1 one-way function. It takes two inputs and compresses
 // them into one output. The inputs and outputs are all of the same size, which
 // is the block size. See [BlockSize].
+// The function must be stateless. Any two calls with the same input must yield the same output.
```
ivokub: Sentence:
> Any two calls with the same input must yield the same output.

duplicates the comment just below

> The output is guaranteed to be the same for the same inputs.

weikengchen:
I want to ask a question about this PR. So now the new code is:
```golang
func (h *merkleDamgardHasher) Sum(b []byte) []byte {
	return append(b, h.state...)
}
```

It doesn't actually do another hash? Just appending b to the string?

weikengchen:
I have a slight concern if any existing use of gnark has been using Sum assuming the old behavior (which could lead to a security issue).

ivokub:
Fair point. But I think this is mitigated by several things:
* it is very recent implementation (from February 2025) and the fix was implemented in June 2025
* we have had no feedback the fix breaking any implementations
* usually, the Sum method is called with `nil` or empty slice, which didn't invoke state update
* previously the implementation was wrong - see the interface description below:
```
// Sum appends the current hash to b and returns the resulting slice.
// It does not change the underlying hash state.
Sum([]byte) []byte
```

So, I don't believe it has caused any security issues in practice. Creating advisories and sending them out is really noisy, and we try to minimize it to keep advisories really high signal.

PR #708, opened 2025-07-09
Feat: Subgroup membership testing on Bandersnatch

yelhousni:
# Description

This PR implements a method for subgroup membership testing on Bandersnatch twisted Edwards elliptic curve. It is based on the paper https://eprint.iacr.org/2022/037.pdf. A slight difference with the paper is that we implement the method directly using twisted Edwards coordinates.

This method can be generalized for all the companion twisted Edwards curves gnark-crypto supports but we need optimized higher residue symbols. As far as I can tell this cannot be as efficient as the binary-GCD-based Legendre symbol. For example Jubjub, we need an octic residue symbol.

### Bandersnatch
Bandersnatch (`B/Fr: -5x^2+y^2=1+dx^2y^2`) has a cofactor 4. Given a point Q purported to be on the prime subgroup of `B`, we need to compute two Tate pairings and check they both are equal to 1. The final exponentiation to `(r-1)/2` is replaced by a Legendre symbol (optimized through Pornin's binary GCD (https://github.com/Consensys/gnark-crypto/pull/704). The Miller functions are just `f1=x-x1` and `f2=x-x2` where `(x1,0)` and `(x2,0)` are of order 2 on the birationally equivalent Weiestrass curve `B'`.

Given `(x_e, x_e)` a point on `B`, `(x, y)` is a point on `B'`, where:
```
      x = ((1+y_e)/(1-y_e) + A/3 ) / B
      A = 2(a+d)/(a-d), B = 4/(a-d)
```
 To avoid inverses we use the fact that `((a/b) / r)_2 = (a * b / r)_2`.
 So `f1` and `f2` are simplified as:
```
      f1 = (1 + y) * (B * (1-y)) and
      f2 = (t0 + t1 * y) * (3B * (1-y))
```
 where:
```
      t0 = 3+A-3B * P'2.X
      t1 = 3-A+3B * P'2.X
```

## Type of change

# How has this been tested?

Added `TestIsInSubGroup`.

The new method is `17x` faster than the naive multiplication by order (on my M1).

# Checklist:


ecc/bls12-381/bandersnatch/point.go
```diff
@@ -168,6 +168,50 @@ func (p *PointAffine) IsOnCurve() bool {
...
+func (p *PointAffine) IsInSubGroup() bool {
+	if p.IsZero() {
+		return true
+	}
+	initOnce.Do(initCurveParams)
+
+	// Given (x_e, x_e) a point on the twisted Edwards curve Ed_{a,d},
+	// (x_w, y_w) is a point on the birationally equivalent short Weierstrass curve W,
+	// where:
+	// 		x_w = ((1+y_e)/(1-y_e) + A/3 ) / B and
```
ThomasPiellard, line 185: y_w is missing in the comment (they are not used in the formulas, but you still can put them and add that only (X_ed, Y_ed) -> X_sw is needed)

ecc/bls12-381/bandersnatch/point.go
```diff
@@ -168,6 +168,50 @@ func (p *PointAffine) IsOnCurve() bool {
...
 }
 
+// IsInSubGroup checks if a point is in the prime subgroup
+// based on https://eprint.iacr.org/2022/037.pdf by D. Koshelev.
+func (p *PointAffine) IsInSubGroup() bool {
+	if p.IsZero() {
+		return true
+	}
+	initOnce.Do(initCurveParams)
+
```
ThomasPiellard, line 181: Unless it is tested beforehand, I think we should check that p is on the curve, since p's x coordinate is not used

ecc/bls12-381/bandersnatch/point.go
```diff
@@ -168,6 +168,50 @@ func (p *PointAffine) IsOnCurve() bool {
...
+	//
+	// To avoid inverses we use the fact that ((a/b) / r)_2 = (a * b / r)_2.
+	// So f_{2,P2}(P) and f_{2,Q2}(P) are simplified as:
+	// 		f1 = (1 + y) * (B * (1-y)) and
+	// 		f2 = (t0 + t1 * y) * (3B * (1-y))
+	// where:
+	// 		t0 = 3+A-3B*Q2.X
+	// 		t1 = 3-A+3B*Q2.X
+
+	var tate1, tate2, temp fr.Element
```
ThomasPiellard, line 214: I don't understand why the formulas are not the same for tate1 and tate2 (with P2.X for tate1 and Q2.X for tate2) ?
Normally you start with the function X_w - P\<i\>.X, and you map it as a function on the twisted Edwards, so for both pairings on the twisted Edwards you should have something like (t\<i\>0 + t\<i\>1 y) \*(3\*B + (1-y)) no ?

ecc/bls12-381/bandersnatch/point.go
```diff
@@ -168,6 +168,50 @@ func (p *PointAffine) IsOnCurve() bool {
...
+	}
+	initOnce.Do(initCurveParams)
+
+	// Given (x_e, x_e) a point on the twisted Edwards curve Ed_{a,d},
+	// (x_w, y_w) is a point on the birationally equivalent short Weierstrass curve W,
+	// where:
+	// 		x_w = ((1+y_e)/(1-y_e) + A/3 ) / B and
+	//  	A = 2(a+d)/(a-d), B = 4/(a-d)
+	//
+	// We need to check that the two tate pairings t_{2,P2}(P) and t_{2,Q2}(P) are 1,
```
ThomasPiellard: Should be renamed P_i (P1 and P2) instead of P2 and Q2, since you use the notation tate1, tate2 for the pairings.

Also, could you provide in the doc the x coordinates of P1 and P2 in the sw curve ?

ecc/bls12-381/bandersnatch/point.go
yelhousni: We can indeed add `IsOnCurve` within the method (as currently for SW) or make it clear in the comments that `IsInSubGroup` checks only subgroup membership. No strong opinion here, for uniformity with SW maybe both inside.

ecc/bls12-381/bandersnatch/point.go
yelhousni: It just happens that for one of the 2-torsion points I found t0=t1=3, but now I cannot find the parameters of the birationally equivalent SW that I used in SageMath ^^. I'll see if I can reconstruct the same one, otherwise I will find a new one and new parameters for verifiability and audit.

ThomasPiellard: ah ok, from the comments it was a bit confusing, I thought there were  2 different formulas^^

ecc/bls12-381/bandersnatch/point.go
ThomasPiellard: no strong opinion either, ok to put `IsOnCurve` inside of `IsInSubGroup`

ThomasPiellard (approved):

PR #719, opened 2025-08-06
Perf: Small Field Montgomery Reduction

Tabaie:
Slight modification to the Montgomery reduction algorithm to combine an addition and comparison into one subtraction.
The classic Montgomery reduction algo goes as follows (source [Wikipedia](https://en.wikipedia.org/wiki/Montgomery_modular_multiplication))

```
function REDC is
    input: Integers R and N with gcd(R, N) = 1,
           Integer N′ in [0, R − 1] such that NN′ ≡ −1 mod R,
           Integer T in the range [0, RN − 1].
    output: Integer S in the range [0, N − 1] such that S ≡ TR−1 mod N

    m ← ((T mod R)N′) mod R
    t ← (T + mN) / R
    if t ≥ N then
        return t − N
    else
        return t
    end if
end function
```

If instead of $N′$ we use $N″ = R-N′$ (due to @DavePearce) we get $m′ ← TN″ \mod R = T(R-N′) \mod R = -TN′ \mod R$.
Since the next multiplication and addition are performed modulo $R²$, after casting (zero-extend) we get $m′ ∈ \\{R-m, -m\\}$, the second case happening when $m = 0 = -m$. Let $t′ ← (T - m′N) / R$. If $m = 0$ then $R|T$ i.e. $T = nR$ for $0 ≤ n < N$, which gives $0 ≤ t′ = n < N$. Otherwise we get $-N ≤ t′ = t - N < N$. In either case $t′ = t \mod N$.

So we end up with:
```
function REDC is
    input: Integers R and N with gcd(R, N) = 1,
           Integer N″ in [0, R − 1] such that NN″ ≡ 1 mod R,
           Integer T in the range [0, RN − 1].
    output: Integer S in the range [0, N − 1] such that S ≡ TR−1 mod N

    m′ ← ((T mod R)N″) mod R
    t′ ← (T - m′N) / R
    if t′ < 0 then
        return t′ + N
    else
        return t′
    end if
end function
```

But this new "range check" is equivalent to whether `T - m′N` produces a borrow out. So may skip the `CMP`.

Benchmarking on an M1 Macbook Pro, currently the gains are a meager %1, but I expect that to rise if/when https://github.com/golang/go/issues/74878 is addressed.

### DIFF
--- a/ecc/bls12-377/fp/element.go
+++ b/ecc/bls12-377/fp/element.go
@@ -932,7 +932,7 @@ func (z *Element) Exp(x Element, k *big.Int) *Element {
 	return z
 }
 
-// rSquare where r is the Montgommery constant
+// rSquare where r is the Montgomery constant
 // see section 2.3.2 of Tolga Acar's thesis
 // https://www.microsoft.com/en-us/research/wp-content/uploads/1998/06/97Acar.pdf
 var rSquare = Element{
@@ -945,7 +945,7 @@ var rSquare = Element{
 }
 
 // toMont converts z to Montgomery form
-// sets and returns z = z * r²
+// sets and returns z = z * r
 func (z *Element) toMont() *Element {
 	return z.Mul(z, &rSquare)
 }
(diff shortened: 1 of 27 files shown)


gbotrel:
That would be a `c7a.8xlarge` I think -- similar CPU target (avx512 support) used on most of our benches

PR #726, opened 2025-08-21
CI: Upgrade linter

Tabaie:
Upgrade `golangci-lint` to `v2.1.6`.
Some changes were made to satisfy the now-stricter `staticcheck` linter.

WARNING: Fixing `ST1008` involved an API change:
```go
func (vector *Vector) AsyncReadFrom(r io.Reader) (int64, error, chan error)
```
is now
```go
func (vector *Vector) AsyncReadFrom(r io.Reader) (int64, chan error, error)
```

### DIFF
--- a/ecc/bls12-377/fp/vector.go
+++ b/ecc/bls12-377/fp/vector.go
@@ -73,12 +73,12 @@ func (vector *Vector) WriteTo(w io.Writer) (int64, error) {
 // It also returns a channel that will be closed when the validation is done.
 // The validation consist of checking that the elements are smaller than the modulus, and
 // converting them to montgomery form.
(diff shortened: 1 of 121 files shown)


ecc/bls12-377/fp/vector.go
```diff
@@ -73,28 +73,28 @@ func (vector *Vector) WriteTo(w io.Writer) (int64, error) {
 // It also returns a channel that will be closed when the validation is done.
 // The validation consist of checking that the elements are smaller than the modulus, and
 // converting them to montgomery form.
-func (vector *Vector) AsyncReadFrom(r io.Reader) (int64, error, chan error) {
+func (vector *Vector) AsyncReadFrom(r io.Reader) (int64, chan error, error) {
```
gbotrel: this API change is going to break upstream usage in gnark / linea

PR #727, opened 2025-08-22
fix: prevent panic on empty vectors in AMD64 vector operations

Fibonacci747:
Fix panic when calling Add/Sub methods on empty vectors in AMD64-optimized
vector implementations across multiple elliptic curve packages.

- Add length check for empty vectors before accessing &a[0] in Add/Sub methods
- Fix InnerProduct validation order to check lengths before empty check
- Ensure consistent behavior with generic implementations
- Apply fixes to all affected curve packages: grumpkin, bn254, stark-curve,
  bls12-381, bls12-377, bls24-315, bls24-317 (both fp and fr packages)
- Add comprehensive test coverage for empty vector operations

This prevents runtime panics that could occur when processing empty vectors
in high-performance cryptographic operations, improving robustness of the
vector math library.

### DIFF
--- a/ecc/bls12-377/fp/vector_test.go
+++ b/ecc/bls12-377/fp/vector_test.go
@@ -74,6 +74,28 @@ func TestVectorEmptyRoundTrip(t *testing.T) {
 	assert.True(reflect.DeepEqual(v3, v2))
 }
 
+func TestVectorEmptyOps(t *testing.T) {
+	assert := require.New(t)
+
+	empty := make(Vector, 0)
+	result := make(Vector, 0)
+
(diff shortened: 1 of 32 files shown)


Fibonacci747:
> I see you already had a commit with tests; similarly for the CI to pass, this needs to go in `field/generator/internal/templates/element/tests_vector.go` to be generated for all implementations

should i close then this PR, or i should just revert changes in these files to make changes in vector_ops_asm.go and tests_vector.go

gbotrel:
we do squash merge so you can work on the same PR (go generate ./... in the root will overwrite your changes anyway -- all these files are generated from templates)

PR #732, opened 2025-08-29
chore: clarify PublicKey.SetBytes uses compressed public key

VolodymyrBg:
Replace misleading x||y description in PublicKey.SetBytes doc.
State that buf must be the compressed public key produced by Bytes() (length sizePublicKey).


ivokub (approved):
Thanks for the pointer. I rewrote the comments completely, they were way off actually.

PR #735, opened 2025-09-02
mpcsetup(generator): use G2Affine in UpdateMonomialsG2 template

Forostovec:
Fix template signature to use []curve.G{{.I}}Affine instead of []curve.G1Affine for UpdateMonomialsG{{.I}}. Ensures UpdateMonomialsG2 is generated with G2Affine across all curves. Aligns with linearCombinationsG{{.I}} and function naming; previous mismatch was a generator typo.


ivokub (approved):
Indeed! Looks good to me. Thanks for the contribution!

PR #752, opened 2025-09-30
feat: add typed field hasher interface in MiMC package

gbotrel:
Adds a interface in `mimc` packages to enable callers to directly work with field elements instead of dancing with bytes (e.g. linea monorepo).

---

> [!NOTE]
> Introduce a typed field-element MiMC hasher with WriteElement/SumElement/SumElements and supporting factories, update internals to reuse buffers, and add tests/templates for all curves.
> 
> - **API**:
>   - Add `FieldHasher` interface with `WriteElement`, `SumElement`, and `SumElements`.
>   - New constructors: `NewFieldHasher` (field elements) and `NewBinaryHasher` (bytes).
> - **Implementation**:
>   - Implement `WriteElement`, `SumElement`, and `SumElements` in `digest` for `bls12-377`, `bls12-381`, `bls24-315`, `bls24-317`, `bn254`, `bw6-633`, `bw6-761`, and `grumpkin`.
>   - Change `Sum` to reset internal buffer with `d.data = d.data[:0]`.
> - **Tests**:
>   - Add `TestFieldHasher` validating equivalence of byte-based and element-based hashing, and `SumElement`/`SumElements` behavior for all curves.
> - **Codegen**:
>   - Update templates (`internal/generator/.../mimc.go.tmpl`, tests) to generate the new API and behavior across curves.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f87fe4a7e6bf6fdf454052c4d40b7f3ddc7969d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ecc/bls12-377/fr/mimc/mimc.go
+++ b/ecc/bls12-377/fr/mimc/mimc.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/crypto/sha3"
 )
 
+// FieldHasher is an interface for a hash function that operates on field elements
 type FieldHasher interface {
 	hash.StateStorer
 
@@ -40,8 +41,6 @@ type FieldHasher interface {
 	// is more efficient.
 	SumElements([]fr.Element) fr.Element
 }
-
-var _ FieldHasher = NewMiMC()
 
 func init() {
 	hash.RegisterHash(hash.MIMC_BLS12_377, func() stdhash.Hash {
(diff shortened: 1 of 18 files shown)


ivokub (commented):
I think the PR makes sense and I think the implementation is good.

I'm just not 100% sure if the interfaces make sense. The problem could be if someone has some kind of switch to initialize the hasher, then because `NewMiMC` in every pacakge now returns different interfaces, then it wouldn't work anymore.

I'd rather instead would define new constructor `NewFieldHasher` which returns `FieldHasher` interface (which we can now simplify to be `Interface {Write(fr.Element) Sum() fr.Element}` and for clarity maybe `NewBinaryHasher` which is the same as `NewMiMC` (and this we keep  around).

This also should cover potential pain-point when people would try to do mixed hashing (i.e. bytes and field elements).

PR #757, opened 2025-10-07
Perf: Optimize `Sqrt` in `Fp2` for all fields

yelhousni:
# Description

Optimise the square root computations in the quadratic extension `Fp2`. Applications: Hash-to-G2 and G2-points decompression.

## Type of change

# How has this been tested?

Current Sqrt tests pass.

- BN254:
```
benchmark             old ns/op     new ns/op     delta
BenchmarkE2Sqrt-8     42712         11773         -72.44%
```

- BLS12-381:
```
benchmark             old ns/op     new ns/op     delta
BenchmarkE2Sqrt-8     118420        28176         -76.21%
```

- BLS12-377:
TODO

# Checklist:


field/generator/internal/templates/element/fixed_exp.go
```diff
@@ -4,6 +4,7 @@ const FixedExp = `
 
 {{- if .SqrtQ3Mod4}}
 	{{expByAddChain "SqrtExp" .SqrtQ3Mod4ExponentData .ElementName}}
+	{{expByAddChain "SqrtExp2" .SqrtQ3Mod4ExponentData2 .ElementName}}
```
gbotrel: Name is not super descriptive, specially now that the method is package public `ExpBySqrtExp2`  ?

field/generator/internal/templates/element/fixed_exp.go
```diff
@@ -15,11 +16,11 @@ const FixedExp = `
 {{- end}}
 
 {{define "expByAddChain name data eName"}}
```
gbotrel, line 18: if we need all these functions public now, maybe we should add a 4th parameter to this template generation function with some extra doc so that the doc is self explanatory

gbotrel (dismissed):
lgtm, just I think the `ExpByXXXX` functions that are now public should have better names / docs, see comments

PR #725, opened 2025-08-20
Feat/sig bounds

ThomasPiellard:
# Description

Ensures s < r/2 in ecdsa

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

See test suite ECDSA

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #767, opened 2025-11-12
Feat: Add P-256 (secp256r1) elliptic curve + ECDSA

yelhousni:
# Description

Add P-256 (secp256r1) elliptic curve (Fr, Fp and G arithmetics + ECDSA).

## Type of change

# How has this been tested?

-

-

# Checklist:

---

> [!NOTE]
> Introduce the P-256 (secp256r1) curve with full field/group ops, MSM, hash-to-curve, and ECDSA, and wire it into the registry, docs, and generators.
> 
> - **New Curve: `secp256r1` (P-256)**
>   - Add full package `ecc/secp256r1`: `fp`/`fr` field arithmetic, G1 (affine/jacobian), multiexp, hash-to-curve (SVDW), ECDSA (sign/verify/recover), (de)serialization, tests/benchmarks.
>   - Implement MSM (batch-affine/jacobian), scalar folding, and hash-to-G1 helpers.
> - **Core Integration**:
>   - Register curve in `ecc`: add `SECP256R1` ID, base/scalar moduli, string mapping; expose generators and curve coefficients.
>   - Update docs (`ecc.md`) to list P-256; refresh package header to include ECDSA, Poseidon2, etc.
>   - Extend codegen config/templates to support `secp256r1` (multiexp, ECDSA, marshal, hash-to-curve).
> - **Misc**:
>   - Tweak secp256k1 hash-to-curve comments/links to RFC 9380.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5cb46b2655858657ff7311ffae760e586c97fe3e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ecc/secp256r1/g1_test.go
+++ b/ecc/secp256r1/g1_test.go
@@ -4,6 +4,7 @@
 package secp256r1
 
 import (
+	"crypto/elliptic"
 	crand "crypto/rand"
 	"fmt"
 	"math/big"
@@ -400,6 +401,37 @@ func TestG1AffineOps(t *testing.T) {
 		},
 		genScalar,
 		genScalar,
+	))
+
+	properties.TestingRun(t, gopter.ConsoleReporter(false))
+}
+
+func TestG1CrossImplementations(t *testing.T) {
+	t.Parallel()
+	parameters := gopter.DefaultTestParameters()
+	if testing.Short() {
+		parameters.MinSuccessfulTests = nbFuzzShort
+	} else {
+		parameters.MinSuccessfulTests = nbFuzz
+	}
+
+	properties := gopter.NewProperties(parameters)
+
+	properties.Property("[SECP256R1] ScalarMultiplication should output the same result as crypto/elliptic", prop.ForAll(
+		func(s fr.Element) bool {
+			curve := elliptic.P256()
+			Gx, Gy := curve.Params().Gx, curve.Params().Gy
+			var S big.Int
+			s.BigInt(&S)
+			Qx, Qy := curve.ScalarMult(Gx, Gy, S.Bytes())
+			var Q G1Affine
+			Q.ScalarMultiplication(&g1GenAff, &S)
+			var qx, qy fp.Element
+			qx.SetBigInt(Qx)
+			qy.SetBigInt(Qy)
+			return Q.X.Equal(&qx) && Q.Y.Equal(&qy)
+		},
+		GenFr(),
 	))
 
 	properties.TestingRun(t, gopter.ConsoleReporter(false))


internal/generator/config/secp256r1.go
```diff
@@ -0,0 +1,28 @@
...
+	G1: Point{
+		CoordType:        "fp.Element",
+		CoordExtDegree:   1,
+		PointName:        "g1",
+		GLV:              false,
+		CofactorCleaning: false,
+		CRange:           []int{4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+	},
+	HashE1: &HashSuiteSvdw{
+		z:  []string{"115792089210356248762697446949407573530086143415290314195533631308867097853948"},
```
ivokub, line 18: I have forgotten - how to we generate these? We don't have hash to curve for secp256r1, do we need it in the conf?

ecc/secp256r1/secp256r1.go
```diff
@@ -0,0 +1,68 @@
...
+// generator of the r-torsion group
+var g1Gen G1Jac
+
+var g1GenAff G1Affine
+
+// point at infinity
+var g1Infinity G1Jac
+
+func init() {
+	aCurveCoeff.SetUint64(3).Neg(&aCurveCoeff)
```
ivokub: Could also use `SetInt64(-3)`?

ecc/secp256r1/marshal.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020-2025 Consensys Software Inc.
+// Licensed under the Apache License, Version 2.0. See the LICENSE file for details.
+
+// FOO
```
ivokub: Whats the comment for?

ecc/secp256r1/marshal.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"errors"
+	"io"
+
+	"github.com/consensys/gnark-crypto/ecc/secp256r1/fp"
+)
+
+// SizeOfG1AffineCompressed represents the size in bytes that a G1Affine need in binary form, compressed
+const SizeOfG1AffineCompressed = 32
```
ivokub: We don't have point compression. So either I think we should remove this constant (this also fixes cursor comment below), or we should add `Bytes()` etc. methods which return uncompressed bytes. Imo in general it would be nice to have `Marshal` etc methods so that we implement `Marshaler` interfaces.

But I think we can leave it out of this PR and do it separately. We have the same issue for secp256k1 as well.

ecc/secp256r1/ecdsa/ecdsa_test.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		b.Fatal(err)
+	}
+	for i := 0; i < b.N; i++ {
+		var recovered PublicKey
+		if err = recovered.RecoverFrom(msg, v, r, s); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
```
ivokub, line 198: What do you think if we would add tests to check compatibility with Go std lib? And if so, then also tests against Wycherproof test vectors? I think we could do it in a future PR.

yelhousni:
@ivokub Added Hash-to-G1 and MSM and fixed review comments 👍

ecc/secp256r1/ecdsa/ecdsa_test.go
yelhousni: Added cross test with crypto/elliptic. I will add tests against Wycherproof either in ecdsa/ or gnark EIP implementation.

yelhousni:
> Thanks for addressing the changes. Looks good now.
> 
> However, the comment by cursor is valid for `v` not being reset in the loops. It is fixed in #725 but that one isn't merged yet as it breaks test generation on gnark side.
> 
> Lets confirm in what order we should merge the PRs.

We can merge this one

ivokub:
> > Thanks for addressing the changes. Looks good now.
> > However, the comment by cursor is valid for `v` not being reset in the loops. It is fixed in #725 but that one isn't merged yet as it breaks test generation on gnark side.
> > Lets confirm in what order we should merge the PRs.
> 
> We can merge this one

You're right. I'll merge this branch and then in #725 will just regenerate to fix the `v`. Then can completely unblock on gnark side.

ecc/secp256r1/ecdsa/ecdsa.go
```diff
@@ -0,0 +1,368 @@
...
+			// set how many times we overflow the scalar field
+			v |= (uint(new(big.Int).Div(r, order).Uint64())) << 1
+			// set if y is even or odd
+			v |= P.Y.BigInt(new(big.Int)).Bit(0)
+
+			r.Mod(r, order)
+			if r.Sign() != 0 {
+				break
+			}
+		}
```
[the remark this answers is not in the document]
ivokub: Will fix in #725

PR #786, opened 2026-01-06
perf: GLV-GLS scalar multiplication in G2

### LINKED ISSUE #78: 4-dim and 8-dim GLS
Currently, gnark-crypto implements 2-dim GLV for all curves and 2-dim GLS for twists over Fp2 (BLS12, BN) and over Fp4 (BLS24). This issue is a placeholder to track implementation of 4-dim GLS (BLS12, BN) and 8-dim GLS (BLS24). This would speedup scalar multiplication on G2 for these curves and potentially multi-scalar-multiplication on G2 (needed for Groth16 in gnark).

yelhousni:
# Description

Implement GLV-GLS scalar multiplication on G2 for bn254, bls12-381 and bls12-377 using two efficient endomorphism phi and psi. 

Fixes #78 

## Type of change

# How has this been tested?

Cross checking double-and-add with mulGLS and mulGLV.

On a AWS c7a.8xlarge`:

- BN254:
```
BenchmarkG2AffineScalarMultiplication
BenchmarkG2AffineScalarMultiplication/method=window
BenchmarkG2AffineScalarMultiplication/method=window-32              7128            167749 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLV
BenchmarkG2AffineScalarMultiplication/method=GLV-32                10000            103615 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLS
BenchmarkG2AffineScalarMultiplication/method=GLS-32                13532             88575 ns/op
```

- BLS12-381:
```
BenchmarkG2AffineScalarMultiplication
BenchmarkG2AffineScalarMultiplication/method=window/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=window/scalarwidth-32                  3568            336292 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLV/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=GLV/scalarwidth-32                     5844            204387 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLS/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=GLS/scalarwidth-32                     6789            174927 ns/op
```

- BLS12-377:
```
BenchmarkG2AffineScalarMultiplication
BenchmarkG2AffineScalarMultiplication/method=window/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=window/scalarwidth-32                  2953            404486 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLV/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=GLV/scalarwidth-32                     5215            228961 ns/op
BenchmarkG2AffineScalarMultiplication/method=GLS/scalarwidth
BenchmarkG2AffineScalarMultiplication/method=GLS/scalarwidth-32                     5811            204052 ns/op
```

# Checklist:

---

> [!NOTE]
> Introduces a 4-dimensional GLV-GLS scalar multiplication path for `G2` on `bn254`, `bls12-381`, and `bls12-377`, leveraging ϕ and ψ endomorphisms for faster large-scalar ops.
> 
> - Adds GLS lattice infra: `Lattice4`, `PrecomputeLattice4`, `SplitScalarFour`, `roundingSigned`, `getVector4`, `det3`; initializes per-curve GLS bases
> - Implements `G2Jac.mulGLS` and routes `ScalarMultiplication`/`ScalarMultiplicationBase` to GLS when beneficial; updates affine wrappers and templates accordingly
> - Adjusts tests/benchmarks to validate GLS vs windowed methods and label runs as GLS; minor cleanup to consistently return from affine scalarmul methods
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45cd07b16c017fdd45b317ab6c83cd77bc7c0abb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ecc/utils.go
+++ b/ecc/utils.go
@@ -260,8 +260,7 @@ func SplitScalar(s *big.Int, l *Lattice) [2]big.Int {
 // SplitScalarFour computes k0,k1,k2,k3 such that
 // k0+k1*lambda1+k2*lambda2+k3*lambda1*lambda2 = s [r].
 // It uses a closest vector approximation in a 4-dimensional lattice.
-// The result is written to res.
-func SplitScalarFour(res *[4]big.Int, s *big.Int, l *Lattice4) {
+func SplitScalarFour(s *big.Int, l *Lattice4) [4]big.Int {
 	var k1, k2, k3, k4 big.Int
 	k1.Mul(s, &l.b1)
 	k2.Mul(s, &l.b2)
@@ -274,11 +273,12 @@ func SplitScalarFour(res *[4]big.Int, s *big.Int, l *Lattice4) {
 	k3.Rsh(&k3, n)
 	k4.Rsh(&k4, n)
 
-	getVector4(res, l, &k1, &k2, &k3, &k4)
+	res := getVector4(l, &k1, &k2, &k3, &k4)
 	res[0].Sub(s, &res[0])
 	res[1].Neg(&res[1])
 	res[2].Neg(&res[2])
 	res[3].Neg(&res[3])
+	return res
 }
 
 // sets res to the closest integer from n/d
@@ -316,17 +316,18 @@ func getVector(l *Lattice, a, b *big.Int) [2]big.Int {
 	return res
 }
 
-// getVector4 computes res = aV1 + bV2 + cV3 + dV4.
-func getVector4(res *[4]big.Int, l *Lattice4, a, b, c, d *big.Int) {
+// getVector4 returns aV1 + bV2 + cV3 + dV4.
+func getVector4(l *Lattice4, a, b, c, d *big.Int) [4]big.Int {
+	var res [4]big.Int
 	coeffs := [4]*big.Int{a, b, c, d}
 	var tmp big.Int
 	for row := 0; row < 4; row++ {
-		res[row].SetUint64(0)
 		for col := 0; col < 4; col++ {
 			tmp.Mul(coeffs[col], &l.V[col][row])
 			res[row].Add(&res[row], &tmp)
 		}
 	}
+	return res
 }
 
 func det3(a00, a01, a02, a10, a11, a12, a20, a21, a22 *big.Int) big.Int {
--- a/ecc/utils_test.go
+++ b/ecc/utils_test.go
@@ -125,8 +125,7 @@ func TestSplittingFour(t *testing.T) {
 	PrecomputeLattice4(&l)
 
 	s.SetString("183927522224640574525727508854836440041603434369820418657580", 10)
-	var k [4]big.Int
-	SplitScalarFour(&k, &s, &l)
+	k := SplitScalarFour(&s, &l)
 
 	var acc, t1, t2, t3 big.Int
 	t1.Mul(&k[1], &lambda1)
@@ -149,8 +148,7 @@ func BenchmarkSplitting256(b *testing.B) {
 	PrecomputeLattice(&r, &lambda, &l)
 	s.SetString("183927522224640574525727508854836440041603434369820418657580", 10)
 
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
+	for b.Loop() {
 		SplitScalar(&s, &l)
 	}
 }
@@ -183,9 +181,7 @@ func BenchmarkSplittingFour256(b *testing.B) {
 
 	s.SetString("183927522224640574525727508854836440041603434369820418657580", 10)
 
-	var k [4]big.Int
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
-		SplitScalarFour(&k, &s, &l)
+	for b.Loop() {
+		SplitScalarFour(&s, &l)
 	}
 }
--- a/ecc/bls12-377/g2.go
+++ b/ecc/bls12-377/g2.go
@@ -858,8 +858,7 @@ func (p *G2Jac) mulGLS(q *G2Jac, s *big.Int) *G2Jac {
 	bases[3].psi(&bases[1])
 
 	// split the scalar, modifies ±q, ϕ(q), ψ(q), ϕ(ψ(q)) accordingly
-	var k [4]big.Int
-	ecc.SplitScalarFour(&k, s, &glsBasis)
+	k := ecc.SplitScalarFour(s, &glsBasis)
 	if k[0].Sign() == -1 {
 		k[0].Neg(&k[0])
 		bases[0].Neg(&bases[0])
--- a/ecc/bls12-381/g2.go
+++ b/ecc/bls12-381/g2.go
@@ -874,8 +874,7 @@ func (p *G2Jac) mulGLS(q *G2Jac, s *big.Int) *G2Jac {
 	bases[3].psi(&bases[1])
 
 	// split the scalar, modifies ±q, ϕ(q), ψ(q), ϕ(ψ(q)) accordingly
-	var k [4]big.Int
-	ecc.SplitScalarFour(&k, s, &glsBasis)
+	k := ecc.SplitScalarFour(s, &glsBasis)
 	if k[0].Sign() == -1 {
 		k[0].Neg(&k[0])
 		bases[0].Neg(&bases[0])
--- a/ecc/bn254/g2.go
+++ b/ecc/bn254/g2.go
@@ -928,8 +928,7 @@ func (p *G2Jac) mulGLS(q *G2Jac, s *big.Int) *G2Jac {
 	bases[3].psi(&bases[1])
 
 	// split the scalar, modifies ±q, ϕ(q), ψ(q), ϕ(ψ(q)) accordingly
-	var k [4]big.Int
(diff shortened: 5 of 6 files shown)


internal/generator/ecc/template/tests/point.go.tmpl
```diff
@@ -620,7 +620,7 @@ func Test{{ $TAffine }}Ops(t *testing.T) {
                 var op1, op2 {{ $TJacobian }}
                 s.BigInt(&r)
                 op1.mulWindowed(&{{.PointName}}Gen, &r)
-                op2.mulGLV(&{{.PointName}}Gen, &r)
+                op2.ScalarMultiplication(&{{.PointName}}Gen, &r)
```
ivokub: I would keep it as is - ScalarMultiplication uses either GLV or Windowed multiplication depending on the scalar length. For random scalars which are usually large this means that `ScalarMultiplication == mulGLV`.

internal/generator/ecc/template/tests/point.go.tmpl
```diff
@@ -595,7 +595,7 @@ func Test{{ $TAffine }}Ops(t *testing.T) {
 
 				r := fr.Modulus()
 				var g {{ $TJacobian }}
-				g.mulGLV(&{{.PointName}}Gen, r)
+				g.ScalarMultiplication(&{{.PointName}}Gen, r)
```
ivokub, line 598: It is fine here, but keep in mind ScalarMultiplication switches between windowed and GLV depending on scalar length.

internal/generator/ecc/template/tests/point.go.tmpl
```diff
@@ -926,7 +926,7 @@ func Benchmark{{ $TJacobian }}ScalarMultiplication(b *testing.B) {
 		b.Run(fmt.Sprintf("method=GLV/scalarwidth=%d", i), func(b *testing.B) {
 			b.ResetTimer()
 			for j := 0; j < b.N; j++ {
-				glv.mulGLV(&{{ .PointName }}Gen, scalar)
+				glv.ScalarMultiplication(&{{ .PointName }}Gen, scalar)
```
ivokub: Keep `mulGLV` here as well. We can switch in the template to call directly either `GLV` or `GLS` depending if G1/G2

ecc/utils_test.go
```diff
@@ -96,6 +96,49 @@ func TestSplitting(t *testing.T) {
...
+	t1.Mul(&k[1], &lambda1)
+	t2.Mul(&k[2], &lambda2)
+	t3.Mul(&k[3], &lambda12)
+	acc.Add(&k[0], &t1).Add(&acc, &t2).Add(&acc, &t3)
+	acc.Sub(&acc, &s)
+	acc.Mod(&acc, &r)
+	if acc.Cmp(&zero) != 0 {
+		t.Fatal("Error split scalar (4D)")
+	}
+}
```
ivokub, line 140: Perhaps also add benchmark? It should be fast, but the lattice precompute adds overhead at program runtime.

ecc/utils.go
```diff
@@ -178,6 +277,43 @@ func getVector(l *Lattice, a, b *big.Int) [2]big.Int {
 	return res
 }
 
+// getVector4 returns aV1 + bV2 + cV3 + dV4.
+func getVector4(l *Lattice4, a, b, c, d *big.Int) [4]big.Int {
```
ivokub: This method can be called every time we do GLS scalar mul. Imo would make sense that the result is already provided as input, then this method is allocate-free.

ecc/utils.go
```diff
@@ -155,6 +218,30 @@ func SplitScalar(s *big.Int, l *Lattice) [2]big.Int {
...
+	k3.Mul(s, &l.b3)
+	k4.Mul(s, &l.b4)
+
+	n := 2 * uint(((l.Det.BitLen()+32)>>6)<<6)
+	k1.Rsh(&k1, n)
+	k2.Rsh(&k2, n)
+	k3.Rsh(&k3, n)
+	k4.Rsh(&k4, n)
+
+	v := getVector4(l, &k1, &k2, &k3, &k4)
```
ivokub: As for getVector4, we can provide the result slice already as input and use it from pool. Perhaps we can implement also `SplitScalarFourNoAlloc` where we provide the result slice as input already. And then in mulGLS we can reuse etc.

Thinking now, I think we can leave it for next PR though, it is self-contained and also applicable for all curves/

ecc/utils.go
```diff
@@ -178,6 +277,43 @@ func getVector(l *Lattice, a, b *big.Int) [2]big.Int {
...
 }
 
+// getVector4 returns aV1 + bV2 + cV3 + dV4.
+func getVector4(l *Lattice4, a, b, c, d *big.Int) [4]big.Int {
+	var res [4]big.Int
+	coeffs := [4]*big.Int{a, b, c, d}
+	for row := 0; row < 4; row++ {
+		res[row].SetUint64(0)
+		for col := 0; col < 4; col++ {
+			var tmp big.Int
```
ivokub: We could initialize tmp once before the loops?

ivokub (changes requested):
Looks good, a few points before merging though:
* in the tests you changed mulGLV to ScalarMultiplication everywhere, but ScalarMultiplication can call mulWindowed when scalar is small. Usually it doesn't happen in tests but imo we had explicit test to test mulGLV output equals mulWindowed output, so we lose the regression test
* on that - we should reconsider the g1ScalarMulChoose and g2ScalarMulChoose constants which switch between GLV/GLS and windowed method. I think it should be smaller now. But it is not clear right now what it should be as due to decomposition overhead its actually less than 64. I think previously it was at around 140-ish bits, but now probably at around 80-90 bits.
- I made a few comments about providing the `*big.Int` slice as an input argument to avoid allocs, but imo we could do it in a future PR (note to self to create an issue for that)

So imo the only necessary change is to consider if we should call `mulGLV`, `mulWindowed` or `ScalarMultiplication` in tests.

ecc/utils.go
```diff
@@ -257,10 +257,11 @@ func SplitScalar(s *big.Int, l *Lattice) [2]big.Int {
 	return v
 }
 
-// SplitScalarFour outputs k0,k1,k2,k3 such that
+// SplitScalarFour computes k0,k1,k2,k3 such that
 // k0+k1*lambda1+k2*lambda2+k3*lambda1*lambda2 = s [r].
 // It uses a closest vector approximation in a 4-dimensional lattice.
-func SplitScalarFour(s *big.Int, l *Lattice4) [4]big.Int {
+// The result is written to res.
+func SplitScalarFour(res *[4]big.Int, s *big.Int, l *Lattice4) {
```
ivokub: You should pass in `[4]*big.Int` instead I think. Usually passing `big.Int` values instead of pointers means you cannot really modify them (think if `big.Int` as `struct {Limbs []uint, sign int}`, so for example when you change sign then it doesn't affect the value. Here I think it works because you pass `*[4]big.Int` with a reference to concrete array, but I think it is a bit subtle.

To be honest, I would keep it as you had before and we should revisit all methods at once in a separate PR, so that we can also modify the usages everywhere else to have zero-alloc scalar splitting.

ecc/utils.go
```diff
@@ -195,14 +234,14 @@ func PrecomputeLattice4(res *Lattice4) {
...
+// SplitScalar computes u,v such that u+vlambda=s[r].
 // The method is to view s as (s,0) in ZxZ, and find a close
 // vector w of (s,0) in <l>, where l is a sub Z-module of
 // ker((a,b) → a+b.λ[r]): then (u,v)=w-(s,0), and
 // u+v.λ=s[r].
+// The result is written to res.
 // cf https://www.iacr.org/archive/crypto2001/21390189.pdf
-func SplitScalar(s *big.Int, l *Lattice) [2]big.Int {
-
+func SplitScalar(res *[2]big.Int, s *big.Int, l *Lattice) {
```
ivokub: Also would keep this as is for now. SplitScalar is a public method, we also depend on it in gnark etc I think so it would be breaking hints.

Lets revisit in another PR with a separate methods `SplitScalarDst` etc.

ecc/utils_test.go
```diff
@@ -148,8 +150,44 @@ func BenchmarkSplitting256(b *testing.B) {
...
+	l.V[3][2].Neg(&lambda1)
+	l.V[3][3].SetUint64(1)
+
+	PrecomputeLattice4(&l)
+
+	s.SetString("183927522224640574525727508854836440041603434369820418657580", 10)
+
+	var k [4]big.Int
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
ivokub: Now its better to use `b.Loop()` pattern for benchmarks. It does automatic timer resets etc. See https://pkg.go.dev/testing#hdr-b_N_style_benchmarks

PR #787, opened 2026-01-06
perf: use NAF decomposition in `mulWindow`

### LINKED ISSUE #132: 2-NAF in G1/G2 scalar multiplication
Currently we use a 2-bit windowed 2-dimensional GLV in G1 and G2 for all the curves. We should use 2-NAF.

yelhousni:
# Description

Use `ecc.NafDecomposition` in `utils.go` in `mulWindowed`.

Fixes #132 

## Type of change

# How has this been tested?

Same tests pass.

e.g. BLS12-377:

```
cpu: AMD EPYC 9R14
BenchmarkG1AffineScalarMultiplication/method=GLV-wNAF/w=5-32     129970        115066        -11.47%
BenchmarkG2AffineScalarMultiplication/method=GLV-wNAF/w=5-32     405302        333270        -17.77%
```

# Checklist:

---

> [!NOTE]
> Implements NAF-based scalar multiplication across all curves and updates codegen.
> 
> - Replaces old 2-bit windowed double-and-add with NAF-based `mulWindowed`, adding `mulWindowedMixed` (affine input) and an affine wrapper; handles zero/negative scalars and reduces scalars mod `fr` where needed
> - Adjusts `ScalarMultiplication`/`ScalarMultiplicationBase` to use new paths and return early, using affine generators (`*GenAff`) in non-GLV cases
> - Minor flow tweaks (e.g., moving `FromAffine` only when needed) and identical changes applied to G1/G2 for bls12-377, bls12-381, bls24-315/317, bn254, bw6-633/761, grumpkin, secp256k1
> - Updates `internal/generator/ecc/template/point.go.tmpl` to generate the new logic
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 833801bde0ba867707fe40bdb3283c3ce760ba76. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ecc/bls12-377/g1.go
+++ b/ecc/bls12-377/g1.go
@@ -657,14 +657,31 @@ func (p *G1Jac) IsInSubGroup() bool {
 // mulWindowed computes a double-and-add scalar multiplication p=[s]q in
 // Jacobian coordinates and using NAF encoding.
 func (p *G1Jac) mulWindowed(q *G1Jac, s *big.Int) *G1Jac {
+	var qAff G1Affine
+	qAff.FromJacobian(q)
+	return p.mulWindowedMixed(&qAff, s)
+}
 
+// mulWindowed computes a double-and-add scalar multiplication p=[s]q in
+// affine coordinates and using NAF encoding.
+func (p *G1Affine) mulWindowed(q *G1Affine, s *big.Int) *G1Affine {
+	var res G1Jac
+	res.mulWindowedMixed(q, s)
+	p.FromJacobian(&res)
+	return p
+}
+
+// mulWindowedMixed computes a double-and-add scalar multiplication p=[s]q
+// where q is in affine coordinates, using NAF encoding.
+func (p *G1Jac) mulWindowedMixed(q *G1Affine, s *big.Int) *G1Jac {
 	if s.Sign() == 0 {
 		p.Set(&g1Infinity)
 		return p
 	}
 	var scalar big.Int
 	scalar.Set(s)
-	if scalar.Sign() < 0 {
+	negScalar := scalar.Sign() < 0
+	if negScalar {
 		scalar.Neg(&scalar)
 	}
 	if scalar.BitLen() > fr.Bits {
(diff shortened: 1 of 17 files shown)


yelhousni:
> nice would be nice if we could have a version with no big int for the naf decomposition to avoid allocs (i.e. if scalar < modulus, use the non-big-int version) ?

The scalar decomposition can output negative integers which would be mod-reduced and the bit-length would increase beyond the |mod|/2 bound.

ecc/bls12-377/g1.go
```diff
@@ -654,35 +654,90 @@ func (p *G1Jac) IsInSubGroup() bool {
...
+	scalar.Set(s)
+	if scalar.Sign() < 0 {
+		scalar.Neg(&scalar)
+	}
+	if scalar.BitLen() > fr.Bits {
+		scalar.Mod(&scalar, fr.Modulus())
 	}
+	var naf [fr.Bits + 1]int8
+	nafLen := ecc.NafDecomposition(&scalar, naf[:])
+	if nafLen == 0 {
```
YaoJGalteland: The `G1Jac.mulWindowed ` version lacks this check. Even without it, the function correctly returns 'infinity' `if nafLen == 0`. This check looks not strictly necessary for correctness, but rather serves as an early-exit optimization to avoid the operations below. I think it's better to align the two versions to ensure a consistency

ecc/bls12-377/g1.go
```diff
@@ -654,35 +654,90 @@ func (p *G1Jac) IsInSubGroup() bool {
 
 }
 
-// mulWindowed computes the 2-bits windowed double-and-add scalar
-// multiplication p=[s]q in Jacobian coordinates.
+// mulWindowed computes a double-and-add scalar multiplication p=[s]q in
+// Jacobian coordinates and using NAF encoding.
 func (p *G1Jac) mulWindowed(q *G1Jac, s *big.Int) *G1Jac {
```
YaoJGalteland, line 659: The implementation of G1Jac.mulWindowed and G1Affine.mulWindowed currently contains significant logic duplication. Since the core windowing algorithm (the main loop and scalar processing) is identical, we should consider abstracting this into a single internal helper function.

This would simplify the codebase and ensure that any future optimizations to the windowed multiplication algorithm only need to be applied in one place.

PR #796, opened 2026-01-22
perf: optimize `Sqrt` in highly 2-adic fields 

### LINKED ISSUE #98: Replace in `Sqrt` Tonelli-Shanks by Sarkar
For BLS12-377 and BLS24-315: https://eprint.iacr.org/2020/1407.pdf

yelhousni:
# Description

This PR updates the field code generator to implement Sarkar’s square‑root algorithm ([ePrint 2020/1407](https://eprint.iacr.org/2020/1407.pdf)) for `q ≡ 1 (mod 4)` fields with 2-adicity `v2(q−1) ≥ 10`. It adds configuration logic to pick optimal Sarkar parameters (k, ℓᵢ) per field, generates the Sarkar helper tables/functions, and wires `Sqrt` to call `SqrtSarkar` while still emitting `SqrtTonelliShanks` for benchmarking and cross‑checks. The test/benchmark templates are updated accordingly, and `initSarkar()` is invoked from generated base files. 
The paper suggests that asymptotically the new method is better for `v2(q−1) ≥ 3`, but gnark‑crypto benchmarks show that for small 2‑adicity (e.g., 6 for `secp256k1/fr`) the optimized Tonelli–Shanks implementation can still be faster due to lower constant factors. Based on those measurements we set the practical activation threshold to `v2(q−1) ≥ 10`. This applies to the following fields:

-   `stark-curve/fp`
-   `bls24-315/fp`
-   `bls12-377`
-   `grumpkin/fp`
-   `<all-curves>/fr`

For small fields the gain small and `Sqrt` are not used anyway so we keep Tonelli-Shanks. For bls12-377 and bls24-315, since they admit SSWU HashToG1 and their `fp` is highly 2-adic, we can use Sarkar's algorithm for `SqrtRatioG1` too.

P.S. I also implemented [Muller's algorithm](http://link.springer.com/article/10.1023/B:DESI.0000015890.44831.e2) and Hamburg's algorithm ([ePrint 2025/2083](https://eprint.iacr.org/2025/2083.pdf)) but Sarkar's yield better results in our cases (when pre-computation is allowed).

Fixes #98 

## Type of change

# How has this been tested?

Cross checks between the Tonelli-Shanks' and Sarkar's implementations.

Some examples on a `c7a.8xlarge`AWS machine:
- BLS12-377 `Fp` (-37.54%):
```
BenchmarkElementSqrt/Sarkar-32             53700             22318 ns/op
BenchmarkElementSqrt/Tonelli-Shanks-32     33580             35734 ns/op
```
HashToG1:
`BenchmarkHashToG1-32     171035        107615        -37.08%
`
- BN254 `Fr` (-24.91%):
```
BenchmarkElementSqrt/Sarkar-32            164080              7311 ns/op
BenchmarkElementSqrt/Tonelli-Shanks-32    123232              9736 ns/op
```

- STARK curve `Fp`(-78%):
```
BenchmarkElementSqrt/Sarkar-8              26479             45240 ns/op
BenchmarkElementSqrt/Tonelli-Shanks-8       5784            206094 ns/op
```

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches core finite-field square-root and hash-to-curve `SqrtRatio` arithmetic across multiple curves; correctness depends on new algorithm/parameter selection despite added cross-check tests.
> 
> **Overview**
> Replaces the default `Sqrt` implementation for `q ≡ 1 (mod 4)` fields with high 2-adicity (threshold `v2(q-1) ≥ 10`) from Tonelli–Shanks to **Sarkar’s algorithm**, adding generated precomputed tables (`g^2^i`), helpers, and a thread-safe `sync.Once` init; the previous Tonelli–Shanks path is retained as `SqrtTonelliShanks` for comparison.
> 
> Extends the generators (`field_config.go`, templates) to auto-pick Sarkar block parameters (`k`, `L[i]`) and wire `Sqrt()` to `SqrtSarkar()` when enabled, and updates SSWU hash-to-curve codegen to use Sarkar-based `SqrtRatio` (with explicit `u==0`/`v==0` handling) for eligible curves.
> 
> Tests/benchmarks are updated to benchmark both algorithms and add a property test asserting `SqrtTonelliShanks` and `SqrtSarkar` agree; benchmarks also now use guaranteed squares as inputs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9aed6ea744d3cb3da8fec2bb31be2a8d5628aaac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/generator/hash_to_curve/template/pkg_sswu.go.tmpl
+++ b/internal/generator/hash_to_curve/template/pkg_sswu.go.tmpl
@@ -203,10 +203,16 @@ func {{$CurveTitle}}SqrtRatio(z *{{$CoordType}}, u *{{$CoordType}}, v *{{$CoordT
 // Computes sqrt(u/v) as sqrt(u*v) / v to avoid explicit division
 // Reference: "On the computation of square roots in finite fields" by Palash Sarkar
 
-	// Handle u = 0 case: sqrt(0/v) = 0, which is a quadratic residue
+	// Handle edge cases to avoid infinite loops in Sarkar's algorithm
+	// u = 0: sqrt(0/v) = 0, which is a quadratic residue
 	if u.IsZero() {
 		z.SetZero()
 		return 0
+	}
+	// v = 0: u/v is undefined; return 0 and flag as non-QR to avoid infinite loop
+	if v.IsZero() {
+		z.SetZero()
+		return 1
 	}
 
 	// Initialize Sarkar precomputed tables (thread-safe, runs once)
--- a/ecc/bls12-377/hash_to_curve/g1.go
+++ b/ecc/bls12-377/hash_to_curve/g1.go
@@ -113,10 +113,16 @@ func G1SqrtRatio(z *fp.Element, u *fp.Element, v *fp.Element) uint64 {
 	// Computes sqrt(u/v) as sqrt(u*v) / v to avoid explicit division
 	// Reference: "On the computation of square roots in finite fields" by Palash Sarkar
 
-	// Handle u = 0 case: sqrt(0/v) = 0, which is a quadratic residue
+	// Handle edge cases to avoid infinite loops in Sarkar's algorithm
+	// u = 0: sqrt(0/v) = 0, which is a quadratic residue
 	if u.IsZero() {
 		z.SetZero()
 		return 0
+	}
+	// v = 0: u/v is undefined; return 0 and flag as non-QR to avoid infinite loop
+	if v.IsZero() {
+		z.SetZero()
+		return 1
 	}
 
 	// Initialize Sarkar precomputed tables (thread-safe, runs once)
--- a/ecc/bls24-315/hash_to_curve/g1.go
+++ b/ecc/bls24-315/hash_to_curve/g1.go
@@ -113,10 +113,16 @@ func G1SqrtRatio(z *fp.Element, u *fp.Element, v *fp.Element) uint64 {
 	// Computes sqrt(u/v) as sqrt(u*v) / v to avoid explicit division
 	// Reference: "On the computation of square roots in finite fields" by Palash Sarkar
 
-	// Handle u = 0 case: sqrt(0/v) = 0, which is a quadratic residue
+	// Handle edge cases to avoid infinite loops in Sarkar's algorithm
+	// u = 0: sqrt(0/v) = 0, which is a quadratic residue
 	if u.IsZero() {
 		z.SetZero()
 		return 0
+	}
+	// v = 0: u/v is undefined; return 0 and flag as non-QR to avoid infinite loop
+	if v.IsZero() {
+		z.SetZero()
+		return 1
 	}
 
 	// Initialize Sarkar precomputed tables (thread-safe, runs once)


gbotrel (changes requested):
So, when replacing in the benchmark:
```
a.SetUint64(4)
	a.Neg(&a)
```
by 
```
a.MustSetRandom()
```

You get quite different results -- can you run that and do a lot of random values to see how it distributes?

internal/generator/field/config/field_config.go
```diff
@@ -300,6 +303,11 @@ func NewFieldConfig(packageName, elementName, modulus string, useAddChain bool)
 			if F.UseAddChain {
 				F.SqrtSMinusOneOver2Data = addchain.GetAddChain(&s)
 			}
+
+			if e >= 10 && !F.F31 && F.PackageName != "goldilocks" {
```
ivokub: I would add a comment there. The conditions (if we don't know) seem quite arbitrary.
```
// We use sarkar when:
// - the field has high 2-adicity. Otherwise, due to smaller constants the Tonelli-Shank is more efficient
// - the field is not a small field (i.e. not F.F31 nor goldilocks)
```
I would also add a better detection for small field to avoid hardcoding goldolocks parameters here.
```
if e >= 10 && !F.F31 && F.NbBits > 64
```

internal/generator/hash_to_curve/template/pkg_sswu.go.tmpl
```diff
@@ -197,7 +197,182 @@ func {{$CurveTitle}}SqrtRatio(z *{{$CoordType}}, u *{{$CoordType}}, v *{{$CoordT
...
+
+	// Check if u/v is QR by checking if (u*v)^m has order dividing 2^(e-1)
+	// Note: (u/v | p) = (u*v | p) since (v^(-1) | p) = (v | p) for Legendre symbols
+	t := xM
+	for i := 0; i < {{.SarkarN}} - 1; i++ {
+		t.Square(&t)
+	}
+	isQNr := !t.IsOne() // t should be ±1; if -1 then u/v is not QR
+	isQNrInt := uint64(0)
+	if isQNr {
```
ivokub, line 242: We could merge this `if isQNr` and the one on line 236 - there are no computations outside of the `if` statement.

internal/generator/hash_to_curve/template/pkg_sswu.go.tmpl
```diff
@@ -197,7 +197,182 @@ func {{$CurveTitle}}SqrtRatio(z *{{$CoordType}}, u *{{$CoordType}}, v *{{$CoordT
...
+		isQNrInt = 1
+	}
+
+	// If not QR, we compute sqrt(Z*u/v) = sqrt(Z*u*v) / v instead
+	// y = Z * x = Z*u*v, and we work with y instead of x
+	if isQNr {
+		// Actually we need to recompute: y = Z*u*v
+		x.Mul(u, v)
```
ivokub: Don't need to compute `x.Mul(u, v)` here as it is already computed on line 211 (and also `u, v` don't change meanwhile)

internal/generator/hash_to_curve/template/pkg_sswu.go.tmpl
```diff
@@ -197,7 +197,182 @@ func {{$CurveTitle}}SqrtRatio(z *{{$CoordType}}, u *{{$CoordType}}, v *{{$CoordT
 
 {{ end }}
 
-{{ if eq (mod .FieldSizeMod256 8) 1 }}// https://www.rfc-editor.org/rfc/rfc9380.html#name-sqrt_ratio-for-any-field
+{{ if eq (mod .FieldSizeMod256 8) 1 }}
+{{ if .SarkarEnabled }}
+// Sarkar's algorithm for SqrtRatio - optimized for high 2-adicity fields (e = {{.SarkarN}})
```
ivokub: Hmm - there is quite a lot of duplication here compared to `fp.SqrtSharkar` -- I get it, here we compute square roots of `z * u/v` and `u/v` in parallel without the expensive division. 

But what if we just try to compute sqrt of `u/v` and if it doesn't exist then of `z * u/v`? We call it only once in the G1 Map To Curve.

Or if it doesn't make sense, then I think it would still be better that the Sharkar sqrt implementations would be somewhat same. Right now it is equivalent, but one version is unrolled, uses inline constant values (i.e. `46` instead of `sarkarN` etc), so really difficult to follow.

internal/generator/config/hash_to_curve.go
```diff
@@ -199,4 +251,12 @@ type HashSuiteInfo struct {
 	Z                []big.Int // z (or zeta) is a quadratic non-residue with //TODO: some extra nice properties, refer to WB19
 	CofactorClearing bool
 	MappingAlgorithm FieldElementToCurvePoint
+
+	// Sarkar parameters for optimized SqrtRatio when 2-adicity e >= 10
```
ivokub, line 247: I think we can leave all Sarkar configuration out of here, no? The information if it is available is already in `field.Field` (passed to `func (suite *HashSuiteSswu) GetInfo(baseField *field.Field, g *Point, name string) HashSuiteInfo`) and also in the current type `HashSuiteInfo.Field.Field`.

And this structure is available in the template.

internal/generator/field/template/element/sqrt.go.tmpl
```diff
@@ -244,6 +244,123 @@ func approximateForLegendre(x *{{.ElementName}}, nBits int) uint64 {
...
+}
+
+// g = nonResidue ^ s
+var sarkarG = {{.ElementName}}{
+	{{- range $i := .SqrtG}}
+	{{$i}},{{end}}
+}
+
+var sarkarGPow [sarkarN]{{.ElementName}}
+var sarkarMinusOne {{.ElementName}}
```
ivokub: We don't have to call it `sarkarMinusOne`. It is just `minusOne`

internal/generator/config/hash_to_curve.go
yelhousni: `computeSarkarBlockSizes` in `hash_to_curve.go` uses a different algorithm (ceiling division targeting block size ~7) than `chooseSarkarParams` in `field_config.go`. They're not directly interchangeable without more investigation.

ivokub: I see now indeed

PR #832, opened 2026-04-15
feat: add Octobear curve and multiset hash on koalabear

yelhousni:
# Description

This PR adds support for the Octobear curve over the degree-8 extension of KoalaBear field and introduces a native multiset hash built on top of it, including a post-quantum vector extension. The target use case is circuit-friendly multiset hashing for computations defined over the KoalaBear field, in particular zkVM memory arguments that will later be verified inside SNARK circuits.

Concretely, this PR:
- adds the new `ecc.OCTOBEAR` curve and the `ecc/octobear` package
- reuses `field/koalabear` for the base field and adds shared `E8` support in `field/koalabear/extensions`
- adds the `octobear` curve arithmetic, scalar field, generator config, and tests
- adds `ecc/octobear/multiset-hash`, implementing three flavors of multiset hash on `octobear`:
  - **classical one-point ECMSH** via the deterministic `y`-increment map (uint16 messages)
  - **vector ECMSH with a linear domain separator** (`N=23` coordinates, `T=128`, `M=2^18`, uint32 messages)
  - **vector ECMSH with a Poseidon2-sponge domain separator** (`N=23` coordinates, `T=256`, width-16 sponge, uint64 messages)
- adds recursive cube-root support over the KoalaBear extension tower used by the Cardano solver
- updates the generator/templates needed to support `octobear` while keeping the rest of the repository green

The one-point construction maps `uint16` messages by searching `y = 256*m + k` for `k < 256`, solving the resulting depressed cubic over `Fp^8`, and accumulating mapped points on `octobear`. The final implementation uses recursive Cardano with binary Lucas sequences at every extension level.

The vector construction keeps `N=23` independently domain-separated ECMSH accumulators. After Shor's algorithm computes discrete logarithms, a collision in the vector digest reduces to a bounded modular linear relation, i.e. a SIS-shaped problem with modulus `r ≈ 2^248` and dimension `N=23` (`23·248 = 5704` bits, matching the Linea KoalaBear LtHash baseline). The linear separator is cheap but leaves a "probably" structurally regular post-Shor matrix; the Poseidon2 separator pays a sponge cost but yields a less structured matrix and is the preferred concrete derivation. Both variants preserve inverse-freeness (`y_i < p/2`) by construction.

## Type of change

# How has this been tested?

The following checks pass locally:

- `go test ./field/koalabear/extensions ./ecc/octobear/...`

Additional validation performed during development:

- exhaustive `Map(uint16)` validation over all `65536` inputs of the one-point variant
- per-coordinate inverse-free invariant (`y_i < p/2`) checked on sampled messages for both vector variants
- homomorphic additivity (`Hash(A ∪ B) == Hash(A) + Hash(B)` componentwise) for both vector variants

Benchmarks were run locally on a MacBook Pro class machine (`darwin/arm64`, Apple M5, 32GB RAM) with:

```bash
go test ./ecc/octobear/multiset-hash -bench . -run '^$'
```

Current results:

- `BenchmarkMap-10`: `25440 ns/op`
- `BenchmarkAccumulatorInsert-10`: `6896592 ns/op` (256 inserts of the one-point variant)
- `BenchmarkHash256-10`: `6802114 ns/op` (one-point hash over 256 messages)
- `BenchmarkMapLinear-10`: `512971 ns/op` (linear vector map for one message → 23 points)
- `BenchmarkMapPoseidon2-10`: `570382 ns/op` (Poseidon2 vector map for one message → 23 points)
- `BenchmarkHashLinear256-10`: `157928583 ns/op` (linear vector hash over 256 messages)
- `BenchmarkHashPoseidon2_256-10`: `160487518 ns/op` (Poseidon2 vector hash over 256 messages)

# Checklist:

---

> [!NOTE]
> **High Risk**
> Introduces a large body of new curve and scalar-field cryptography (including cube roots and multiset hashes) intended for security-sensitive zkVM arguments, with limited audit coverage called out in generated code.
> 
> **Overview**
> This PR **registers the Octobear curve** (`ecc.OCTOBEAR`) and wires scalar/base field metadata in `ecc/ecc_field.go`, with **misspell** configured to ignore the name `octobear`.
> 
> It adds the **`ecc/octobear` tree**: `fp` is a thin alias over **`field/koalabear`** so imports match other curves; **`fr`** is a full generated ~248-bit scalar field (Montgomery arithmetic, amd64/arm64 asm, vectors, **Tonelli–Shanks `Sqrt` and cubic-root `Cbrt`** with dedicated exp chains and tests). The curve package targets **𝔽_p⁸** over KoalaBear for circuit-friendly use (per package docs), including **multiset-hash** variants described in the PR (classical ECMSH, vector ECMSH with linear or Poseidon2 domain separation).
> 
> **Review focus:** correctness of new field/curve parameters, `Cbrt`/extension math used by hashing, and performance/security assumptions of the multiset-hash constructions—not re-auditing every generated `fr` line unless behavior diverges from other curves.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6b0b478eda65300e89ed275840ae7b127cc0f80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/generator/main.go
```diff
@@ -163,6 +166,13 @@ func main() {
 				assertNoError(field.GenerateFF(conf.Fr, outputDir, frOpts...))
 			}
 
+			// preserve the checked-in kb8 ECC package; the shared ECC generator remains
+			// master-neutral for existing curves, while kb8 keeps its hand-maintained
+			// field-wrapper, point, marshal, and multiexp files.
+			if conf.Equal(config.KB8) {
```
ivokub: I think we could use the `NoEcc=true` configuration in the curve parameter? I'll try if it works.

ecc/octobear/fp/fp.go
```diff
@@ -0,0 +1,42 @@
+package fp
```
ivokub, line 6: I think for long-term maintainability, we should perhaps update the generator to indicate that fp is an existing package and we can point to its location so that this file would be generated as well.

gbotrel:
Review pass from 2026-05-07.

Findings:
- Blocking: staticcheck CI is correctly failing because the generated code is stale. The Actions log reports that `go generate` leaves the repository dirty, with a diff in `ecc/kb8/fr/element.go` (the generated exponentiation helpers differ from the committed file). Please run the generator and commit the resulting generated file before this PR is used as the dependency for gnark #1757.

Verification run locally:
- `go test ./field/koalabear/extensions ./ecc/kb8/...`

ecc/bn254/multiexp_glv.go
```diff
@@ -0,0 +1,1076 @@
...
+func (ctx *g1GLVStripeBatchAffine) resetStripe() {
+	for _, bucketID := range ctx.touched {
+		ctx.buckets[bucketID].SetInfinity()
+		ctx.bucketsJE[bucketID] = g1JacExtended{}
+		ctx.touchedMarks[bucketID] = false
+	}
+	ctx.touched = ctx.touched[:0]
+	ctx.qID = 0
+	ctx.resetBatch()
+}
```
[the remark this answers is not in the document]
yelhousni: Pippenger prefix-sum reduction; total += runningSum on untouched buckets is algorithmically required for the dense-path identity total = Σ k·bucket[k] = Σ runningSum[k]. A sparsity-aware path exists but is a separate optimization tracked outside this PR.

ecc/bn254/multiexp_glv.go
```diff
@@ -0,0 +1,1076 @@
...
+	ctx.flushQueue()
+	for k := len(ctx.buckets) - 1; k >= 0; k-- {
+		runningSum.addMixed(&ctx.buckets[k])
+		if !ctx.bucketsJE[k].IsInfinity() {
+			runningSum.add(&ctx.bucketsJE[k])
+		}
+		total.add(runningSum)
+	}
+	ctx.resetStripe()
+}
```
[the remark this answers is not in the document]
yelhousni: ditto

internal/generator/field/template/extensions/e8.go.tmpl
```diff
@@ -0,0 +1,304 @@
...
+		Sub(&c, &a).
+		Sub(&c, &b)
+	z.C1.Set(&c)
+	b.MulByNonResidue(&b)
+	z.C0.Add(&a, &b)
+	return z
+}
+
+// Square sets z=x*x in E8 and returns z.
+func (z *E8) Square(x *E8) *E8 {
```
ivokub, line 229: Isn't there more optimal way for Square? There is for E4 for example?

internal/generator/field/generator_extensions.go
```diff
@@ -19,11 +19,21 @@ func generateExtensions(F *config.Field, outputDir string) error {
...
 	entries_ext2 := []bavard.Entry{
 		{File: filepath.Join(outputDir, "doc.go"), Templates: []string{"doc.go.tmpl"}},
 		{File: filepath.Join(outputDir, "utils.go"), Templates: []string{"utils.go.tmpl"}},
-		{File: filepath.Join(outputDir, "e2.go"), Templates: []string{"e2.go.tmpl"}},
-		{File: filepath.Join(outputDir, "e2_test.go"), Templates: []string{"e2_test.go.tmpl"}},
+	}
+	// koalabear has a hand-maintained recursive cube-root stack in its extension files.
+	// The generic extension templates don't reproduce it yet, so generation only refreshes
+	// the shared support files and leaves e2/e4/e8 implementation files untouched.
+	if !isKoalaBear {
```
ivokub: Imo we shouldn't add this exception for Koalabear. What we could do instead is to have the cbrt implementation in a separate file e2_cbrt.go what we hand-maintain (i.e. the Cbrt method and the constants), but the rest of e2/e4 would still be in code-generated files.

internal/generator/field/generator_field.go
```diff
@@ -153,7 +153,9 @@ func generateField(F *config.Field, outputDir, asmDirIncludePath, hashArm64, has
 
 	g.Go(generate("vector_purego.go", []string{"element/vectoropspurego.go.tmpl"}, withBuildTag(pureGoVectorBuildTag)))
 
-	if F.UseAddChain {
+	// koalabear keeps custom fixed-exponent helpers alongside its recursive extension cube-root stack.
+	// The generic field template doesn't model those helpers yet, so generation must leave element_exp.go intact.
+	if F.UseAddChain && F.PackageName != "koalabear" {
```
ivokub: I'd also avoid the exception here. We could define in the field conf CbrtQ2Mod9 and then have the exp code generated?

internal/generator/fieldwrapper/template/fp.go.tmpl
```diff
@@ -0,0 +1,40 @@
+import (
```
ivokub, line 1: Template itself is fine, I'd add comments (even though the method names are self-explanatory). But definitely package doc would be useful saying that this aliases the destination package but provide a common interface for ecc/KB8

internal/generator/config/curve.go
```diff
@@ -14,6 +14,7 @@ type Curve struct {
 	EnumID       string
 	FpModulus    string
 	FrModulus    string
+	ExistingFp   ExistingFieldPackage
```
ivokub: Imo comment for this field would be useful - when it should be set for example.

internal/generator/main.go
```diff
@@ -92,6 +93,12 @@ func main() {
 	// files with the "DO NOT EDIT" header are removed; hand-written files
 	// (without this header) are preserved.
 	for _, conf := range config.Curves {
+		if conf.Equal(config.KB8) {
```
ivokub: I am not sure we need this exceptional case handling -- the function `cleanGeneratedFiles` checks if there is the `DO NOT EDIT` header in the files and deletes those. To avoid deleting hand-written implementation it is sufficient that we remove the header in those files.

ecc/bn254/multiexp_glv.go
```diff
@@ -0,0 +1,1077 @@
+// Copyright 2020-2026 Consensys Software Inc.
+// Licensed under the Apache License, Version 2.0. See the LICENSE file for details.
+
+package bn254
```
ivokub: WHy are these files included in this PR? It should be about KB8 only?

ivokub (changes requested):
Sending first round of review. This is mostly about the code generation part.

I think for some implementation having hand-written files is fine, but in that case they should explicilty not have the "DO NOT EDIT" header, this breaks the automatic deletion logic. And I think we can still even now have more code-generation coverage and avoid switching in the generator function on concrete curves (and rather use attribute-base generation logic). Please see the inline comments.

I will do the curve implementation and multiset hash in another round. But it would be easier if we would have more code-generated parts (for the ECC even), so that I know everything matches the templates. What was the block change why files in `ecc/kb8` had to be hand-written, not generated? Imo seems quite similar to other curves?

ecc/bn254/multiexp_glv.go
yelhousni: oh  that some dirty state from another branch. thanks for catching!

field/koalabear/extensions/e2_cbrt.go
```diff
@@ -0,0 +1,190 @@
...
+	gamma0.Mul(&wa1, &k).Mul(&gamma0, &cbrtFpThree)
+	gamma1.Mul(&wa0, &k)
+
+	var mInv fr.Element
+	mInv.Square(&m).Mul(&mInv, &normInv)
+
+	var t1, t2 fr.Element
+	t1.Mul(&x.A0, &gamma0)
+	t2.Mul(&x.A1, &gamma1).Mul(&t2, &cbrtFpThree)
+	z.A0.Sub(&t1, &t2).Mul(&z.A0, &mInv)
```
ivokub: This Cbrt path is not receiver-safe. In the general E2 path, `z.A0` is assigned here before line 100 still reads `x.A0`, so `x.Cbrt(&x)` can corrupt the input and return `nil` for a valid cube. Please compute the candidate into a local `y E2`, verify that, and only assign to `z` at the end. The E4/E8 subfield branches inherit this issue when they call nested Cbrt with receiver/input aliases.

ecc/octobear/multiset-hash/vector_multiset_hash_poseidon2.go
```diff
@@ -0,0 +1,211 @@
...
+	)
+
+	// Absorb (domainTag, msg) into the rate part (state[0:pqSqueezeRate]).
+	// The 8-byte tag is split into two 32-bit big-endian halves and the
+	// 64-bit msg into two more halves, filling state[0..3]. state[4..7]
+	// and the capacity state[8..15] stay zero.
+	state[0].SetUint64(uint64(binary.BigEndian.Uint32(pqDomainTag[0:4])))
+	state[1].SetUint64(uint64(binary.BigEndian.Uint32(pqDomainTag[4:8])))
+	state[2].SetUint64(msg & 0xFFFFFFFF)
+	state[3].SetUint64(msg >> 32)
```
ivokub: This encoding is not injective for the public `uint64` message domain. `Element.SetUint64` reduces each 32-bit chunk modulo `p`, and `p = 2130706433 < 2^32`, so `msg = 0` and `msg = p` absorb the same low limb and the same high limb. That gives `MapPoseidon2(0) == MapPoseidon2(p)`, hence a trivial single-element multiset collision. Please encode `uint64` injectively, for example with four 16-bit field elements or three base-`p` limbs, and add regression tests for this boundary.

ecc/octobear/multiset-hash/doc.go
```diff
@@ -0,0 +1,26 @@
...
+//   - Poseidon2-sponge vector ECMSH (Poseidon2Accumulator, HashPoseidon2,
+//     MapPoseidon2): a digest of N = 23 curve points. The N ordinates are
+//     derived by absorbing (domain tag, msg) into a width-16 Poseidon2
+//     sponge (rate 8, 3 squeeze permutations) and range-reducing each
+//     output into [0, floor(p/(2T))) with T = 256.
+//
+// The two vector variants are post-quantum candidates: under Shor's
+// algorithm a collision becomes a bounded modular linear relation, i.e. a
+// SIS-shaped problem with modulus r ~ 2^248 and dimension N = 23
+// (5704 SIS-volume bits, matching the Linea KoalaBear LtHash baseline).
```
ivokub, line 26: I cannot fully check this post-quantum security claim without the referenced construction/paper. The docs should cite the concrete `ECMSH_PQ` or LtHash-style reference and spell out the bounded-multiset parameters used by the SIS estimate, in particular coefficient/cardinality bounds and the target security level. The `5704` SIS-volume-bit number is useful context, but by itself it is not enough to review the security claim.

ivokub (changes requested):
I now extended the review to the Octobear curve as well. This looks good. I tried locally implementing code-generation for octobear curve and it seems to work. But it is a bit bigger change and imo we should do it in a separate PR after this PR is done.

I also started looking at the multi-set hash. There are a few quickly fixable issues. However, to estimate the security level, it would be better to have concrete references to the paper (right now we just referene as ECMSH_PQ without concrete link), right now I'm not sure about the claim :)

After the small fixes and paper references I think the PR is ready to merge (I'll run another round of review though to be sure)

ecc/octobear/multiset-hash/multiset_hash.go
```diff
@@ -0,0 +1,95 @@
...
+
+var errMapFailure = errors.New("octobear multiset hash: failed to map message after 256 y-increments")
+
+// Accumulator stores an additive multiset hash state in affine coordinates.
+type Accumulator struct {
+	sum octobear.G1Affine
+}
+
+// NewAccumulator returns a zero accumulator.
+func NewAccumulator() Accumulator {
```
ivokub, line 19: Perhaps should be a pointer? Then we can pass around the accumulator between different owners which modify the same thing. But in this case we would most probably need to add some mutex locking.

Right now when we receive the `acc := NewAccumulator(); utils.Parallel(..., func() error {accumulat(accumulator)});` then every parallel instance would get an initial copy. Or we can leave it as is and just document more that passing by value copies.

ecc/octobear/multiset-hash/multiset_hash.go
```diff
@@ -0,0 +1,95 @@
...
+			return octobear.G1Affine{}, err
+		}
+	}
+	return acc.Digest(), nil
+}
+
+// Map deterministically maps msg to a point on octobear using the y-increment method.
+// It returns the mapped point and the first offset k in [0, 255] such that
+// y = msg*256 + k yields a point (x, y) on octobear.
+func Map(msg uint16) (octobear.G1Affine, uint8, error) {
```
ivokub, line 70: Could reuse mapAtBase here.
```go
return mapAtBase(baseY, tweakBound, b)
```

ecc/octobear/multiset-hash/multiset_hash.go
```diff
@@ -0,0 +1,95 @@
+package multisethash
+
+import (
+	"errors"
+
+	"github.com/consensys/gnark-crypto/ecc/octobear"
+	"github.com/consensys/gnark-crypto/field/koalabear/extensions"
+)
+
+const tweakBound = 256
```
ivokub, line 9: The const is shadower in mapAtBase. Perhaps we should rename it there? Otherwise when refactoring etc could have some odd results.

ecc/octobear/multiset-hash/cardano_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestE16CbrtZero(t *testing.T) {
+	var zero, got e16
+	require.NotNil(t, got.Cbrt(&zero))
+	require.True(t, got.A0.IsZero())
+	require.True(t, got.A1.IsZero())
+}
+
+func TestDepressedCubicRootFindsValidRoot(t *testing.T) {
```
ivokub, line 70: The Cardano solver structure looks consistent with the secp256r1 one, and I do not see a correctness issue in the implementation. I would still add explicit branch tests here: one for the repeated-root cases (`c = ±2` for `x^3 - 3x + c = 0`), one that exercises the square-discriminant/base-field path, and one that exercises the non-square-discriminant/quadratic-extension path. The current randomized test verifies returned roots, but it does not make it obvious that all Cardano branches are covered.

ivokub (dismissed):
Checked now multiset-hash as well. Imo the implementation is clear, but I'm still a bit unsure about the soundness level.

Left a few more comments about potentially making the accumulator more parallelizable. See if it makes sense.

And I still think we should code-generate octobear. But perhaps when it is a bit more settled as a separate PR :) My local experiments show that its doable, the issue is mostly about serialization (due to E8 case handling we need to add) and some different formulas for EC imo.

ecc/octobear/multiset-hash/multiset_hash.go
yelhousni: I think mutex-shared state would defeat the homomorphic combine pattern

yelhousni:
> Checked now multiset-hash as well. Imo the implementation is clear, but I'm still a bit unsure about the soundness level.
> 
> Left a few more comments about potentially making the accumulator more parallelizable. See if it makes sense.
> 
> And I still think we should code-generate octobear. But perhaps when it is a bit more settled as a separate PR :) My local experiments show that its doable, the issue is mostly about serialization (due to E8 case handling we need to add) and some different formulas for EC imo.

Created an issue for the code-generation #846.

field/koalabear/extensions/e2_cbrt.go
```diff
@@ -0,0 +1,191 @@
...
+// It returns nil if x is not a cubic residue.
+//
+// The candidate is computed into a local E2 and only written to z at the
+// end, so x.Cbrt(&x) is safe (does not corrupt x mid-computation).
+func (z *E2) Cbrt(x *E2) *E2 {
+	if x.A1.IsZero() {
+		z.A0.Cbrt(&x.A0)
+		z.A1.SetZero()
+		return z
+	}
```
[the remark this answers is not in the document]
yelhousni: cbrt on koalabear is a bijection so will always return a result

ivokub (approved):

PR #833, opened 2026-04-20
feat: add subgroup membership tests for all `twistededwards` curves

yelhousni:
# Description

This PR adds divide-and-pair subgroup membership tests (SMT) for all twisted Edwards curves in `gnark-crypto`, following the https://eprint.iacr.org/2026/749 and https://hackmd.io/@yelhousni/divide-and-pair. The implementation uses generated Tate-style subgroup checks together with addchain-generated exponentiations for quartic and octic residue tests.

The generator was extended to derive the relevant subgroup structure metadata from the curve parameters, emit the appropriate subgroup-check code, and generate tests and benchmarks for all supported twisted Edwards curves. This also adds benchmarks comparing `IsInSubGroup()` against multiplication by the subgroup order, so the new method can be measured directly against the previous baseline on each curve.

## Type of change

# How has this been tested?

```bash
go test ./ecc/bn254/twistededwards \
  ./ecc/bls12-377/twistededwards \
  ./ecc/bls12-381/twistededwards \
  ./ecc/bls12-381/bandersnatch \
  ./ecc/bls24-315/twistededwards \
  ./ecc/bls24-317/twistededwards \
  ./ecc/bw6-633/twistededwards \
  ./ecc/bw6-761/twistededwards
```

The tests check that:
- the identity point is accepted
- random scalar multiples of the base point are accepted
- the rational `2`-torsion point `(0,-1)` is rejected

Benchmarks were run with:

```bash
go test -run '^$' -bench IsInSubGroup \
  ./ecc/bn254/twistededwards \
  ./ecc/bls12-377/twistededwards \
  ./ecc/bls12-381/twistededwards \
  ./ecc/bls12-381/bandersnatch \
  ./ecc/bls24-315/twistededwards \
  ./ecc/bls24-317/twistededwards \
  ./ecc/bw6-633/twistededwards \
  ./ecc/bw6-761/twistededwards
```
- [x] `BenchmarkIsInSubGroup`, on Apple M5, macOS arm64

Representative results on Apple M5 / macOS arm64:

- `bn254`: `is_in_subgroup` ~`5.4µs` vs `mul_by_order` ~`29.8µs`
- `bls12-381/twistededwards`: `is_in_subgroup` ~`5.2µs` vs `mul_by_order` ~`29.9µs`
- `bandersnatch`: `is_in_subgroup` ~`1.3µs` vs `mul_by_order` ~`31.9µs`
- `bw6-761`: `is_in_subgroup` ~`11.1µs` vs `mul_by_order` ~`80.9µs`

# Checklist:

---

> [!NOTE]
> **High Risk**
> Incorrect subgroup membership logic would accept invalid curve points in cryptographic paths; the change is large, curve-specific generated math with new exponentiation chains.
> 
> **Overview**
> Adds **divide-and-pair** prime-subgroup checks for every supported `twistededwards` package via new generated `subgroup.go` files: Edwards points map to a short Weierstrass model, Miller-style line products are raised with **addchain**-generated `expByQuarticExp` / `expByOcticExp`, and cofactor-8 curves also precompute rational 8-torsion data at init (`subgroupInitOnce`).
> 
> The **edwards generator** now picks the check from curve metadata (Legendre of `a·d` for split cofactor-4 / Koshelev vs non-split cofactor 4 or 8), emits `subgroup.go` from a new template, and wires **bandersnatch** split-2-torsion constants in config. **Tests** add agreement with scalar multiplication by subgroup order, rejection of torsion cosets, and **benchmarks** compare `IsInSubGroup` to `mul_by_order`. Minor generator fixes: addchain cache path via `runtime.Caller`, odd modulus validation for fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ddca999dac9150fdedb43584dd810da106a66ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ecc/bls12-377/twistededwards/subgroup.go
```diff
@@ -0,0 +1,348 @@
+// Copyright 2020-2026 Consensys Software Inc.
+// Licensed under the Apache License, Version 2.0. See the LICENSE file for details.
+
+// Code generated by consensys/gnark-crypto DO NOT EDIT
+
+package twistededwards
+
+import "github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
+
+type weierstrassPointAffine struct {
```
ivokub, line 13: Why do we define another affine point struct here? Cannot we just use the PointAffine definition?

ecc/bls12-377/twistededwards/subgroup.go
```diff
@@ -0,0 +1,348 @@
...
+	z.Mul(z, &t0)
+
+	for range 45 {
+		z.Square(z)
+	}
+
+	return z
+}
+
+// IsInSubGroup checks if a point is in the prime subgroup.
```
ivokub, line 335: Could we add references to the papers? Also perhaps inline we could have more comments about what is happening and how it works.

ecc/bls12-377/twistededwards/subgroup.go
ivokub: Even if there is difference between Edwards and Weierstrass forms, then as we only use it internally then perhaps it is fine to keep the same type definition? We would just have to indicate in the variable name "weierstrassAffine..." etc

internal/generator/edwards/template/curve.go.tmpl
```diff
@@ -74,13 +73,11 @@ func initCurveParams() {
 
 	curveParams.Base.X.SetString("{{.BaseX}}")
 	curveParams.Base.Y.SetString("{{.BaseY}}")
-	{{- if eq .Cofactor "4"}}
-	{{- if eq .Name "bls12-381"}}
+	{{- if and (eq .Cofactor "4") .Split2Torsion}}
 	curveParams.t0.SetString("44968234042453258989421494579017642355260750649112422763795205757285533011434")
```
ivokub: Why are the constants defined here? Aren't they curve-specific and should be defined in the configuration from which we generate?

internal/generator/edwards/generate.go
```diff
@@ -12,11 +13,18 @@ import (
...
 		{File: filepath.Join(baseDir, "point.go"), Templates: []string{"point.go.tmpl"}},
+		{File: filepath.Join(baseDir, "subgroup.go"), Templates: []string{"subgroup.go.tmpl"}},
 		{File: filepath.Join(baseDir, "point_test.go"), Templates: []string{"tests/point.go.tmpl"}},
 		{File: filepath.Join(baseDir, "doc.go"), Templates: []string{"doc.go.tmpl"}},
 		{File: filepath.Join(baseDir, "curve.go"), Templates: []string{"curve.go.tmpl"}},
 	}
 
 	edwardsGen := common.NewDefaultGenerator(template.FS)
-	return edwardsGen.Generate(conf, conf.Package, "", "", entries...)
+	funcs := common.Funcs()
```
ivokub: GenerateWithOptions already sets the default functions, we would only need to provide the additional functions. A la 
```diff
 import (
        "path/filepath"
+       stdtemplate "text/template"
 
        "github.com/consensys/bavard"
        "github.com/consensys/gnark-crypto/internal/generator/addchain"
@@ -20,7 +21,7 @@ func Generate(conf config.TwistedEdwardsCurve, baseDir string, gen *common.Gener
        }
 
        edwardsGen := common.NewDefaultGenerator(template.FS)
-       funcs := common.Funcs()
+       funcs := make(stdtemplate.FuncMap)
        for _, f := range addchain.Functions {
                funcs[f.Name] = f.Func
        }
```

internal/generator/edwards/template/curve.go.tmpl
```diff
@@ -16,11 +16,9 @@ type CurveParams struct {
 	Cofactor fr.Element
 	Order    big.Int
 	Base     PointAffine
-	{{- if eq .Cofactor "4"}}
-	{{- if eq .Name "bls12-381"}}
+	{{- if and (eq .Cofactor "4") .Split2Torsion}}
```
ivokub: So what I have tried with code generation is to have templates and code generation conditionals be property-based. So for example if we define the method `SupportsSplit2Torsion()` etc on the `TwistedEdwardsCurve` in internal/generator/config/curve.go, then here we just set `{{- if SupportsSplit2Torsion }}` and it is then called at code generation time.

This way we keep the conditions in a single file close to the curve parameter definition and if we change it there it automatically covers all templates etc we have.

ivokub (changes requested):
In general looks good, but I'd make it a bit more documented as the subgroup check uses generally not known algorithm :) Helps later reviews to have references.

Also, please consider refactoring some constants from the templates to configuration and perhaps also make template generation property-based.

gbotrel:
Review pass from 2026-05-07.

Findings:
- Blocking: hosted `test` CI is red. The failing package/test reported by the Actions log is `github.com/consensys/gnark-crypto/internal/generator/field/config` / `TestIntToMont` on the Go 1.25.x job. I could not reproduce it locally on go1.26.2 with either `go test ./internal/generator/field/config -run TestIntToMont -v` or `go test ./internal/generator/field/config -run TestIntToMont -count=2 -timeout 60s -v`, but the PR should not merge with the required test job red. Please rerun with better verbose output or investigate whether the property generator is exposing a Go-version/platform-sensitive case.

Verification run locally:
- `go test ./ecc/bn254/twistededwards ./ecc/bls12-377/twistededwards ./ecc/bls12-381/twistededwards ./ecc/bls12-381/bandersnatch ./ecc/bls24-315/twistededwards ./ecc/bls24-317/twistededwards ./ecc/bw6-633/twistededwards ./ecc/bw6-761/twistededwards`
- `go test ./internal/generator/field/config -run TestIntToMont -v`
- `go test ./internal/generator/field/config -run TestIntToMont -count=2 -timeout 60s -v`

ecc/bls12-377/twistededwards/subgroup.go
yelhousni: I kept a separate type intentionally. The coordinates are structurally the same as PointAffine, but they live on the birational short-Weierstrass model, while PointAffine methods implement Edwards formulas. Reusing PointAffine would make it easy to accidentally call Edwards Add, Double, etc. on Weierstrass coordinates. The local type has no methods and keeps that model boundary explicit.

internal/generator/edwards/template/curve.go.tmpl
yelhousni: Agreed. I moved the split-2-torsion constants out of the template and into the curve config for bandersnatch, then gated generation through an explicit HasSplit2TorsionSubgroupCheck() predicate. This avoids silently applying bandersnatch-specific constants to a future cofactor-4 split-2 curve.

internal/generator/edwards/generate.go
yelhousni: Good point. I changed this to pass only the additional addchain template funcs via a fresh template.FuncMap; GenerateWithOptions already installs the common funcs.

internal/generator/edwards/template/curve.go.tmpl
yelhousni: Agreed. I added methods on TwistedEdwardsCurve such as HasSplit2TorsionSubgroupCheck, HasNonSplitSubgroupCheck, HasOcticSubgroupCheck, and HasSubgroupCheck, and updated the templates to use those instead of open-coded conditionals.

.gitignore
```diff
@@ -32,6 +32,8 @@ internal/tests/integration/
 field/internal/dev.go 
 field/internal/dev/**
 field/generator/addchain/**
+internal/generator/field/addchain/
```
ivokub: You have added into gitignore, but the files are checked into PR. Should we remove from the tree?

Do we need the `/addchain/` entry in gitignore?

ivokub (commented):
Looks good. Before approval just to confirm if we wanted to have the files in `internal/generator/addchain` to be included in the PR?

.gitignore
yelhousni: Good catch, these shouldn't be in the PR. I'll drop them and fix the .gitignore pattern.

ivokub:
@yelhousni - is the last commit correct? It drops the gitignore entry for internal addchain, but the cached files are still around https://github.com/Consensys/gnark-crypto/pull/833/changes#diff-02ba2ce3b66ef8146ee54bc46b614026a1b54783c7782501b1d79e9840b29043.

yelhousni:
> @yelhousni - is the last commit correct? It drops the gitignore entry for internal addchain, but the cached files are still around https://github.com/Consensys/gnark-crypto/pull/833/changes#diff-02ba2ce3b66ef8146ee54bc46b614026a1b54783c7782501b1d79e9840b29043.

The cache at internal/generator/addchain/ has been intentionally committed since 0020165e4 (Nov 2021, "deterministic addchain generation on CI") and this PR followed the same pattern. The two .gitignore lines I removed (internal/generator/field/addchain/, /addchain/) point at paths that no current code writes — the latter was only ever produced when a generator ran from the repo root, which the runtime.Caller(0) change in this commit prevents.

ivokub (approved):
Thanks for the clarification!

PR #841, opened 2026-05-18
fix(amd64): avoid 4-byte overread in innerProdVec AVX-512 path

gbotrel:
## Summary

- `VPMULUDQ.BCST` broadcasts an `m64` (8 bytes), so `MAC(28(R13), ...)` inside the AVX-512 `innerProdVec` inner loop reads bytes [28..35] of each 32-byte element — 4 bytes past the element boundary.
- For interior elements those bytes harmlessly belong to the next element; for the **last** element of the slice they land past the allocation and SIGSEGV when the slice ends exactly at a page boundary (e.g. `n = 262144` for a 4-word field on Linux amd64).
- Fix: replace only the offset-28 MAC with a 4-byte `VPBROADCASTD` followed by a register-form `VPMULUDQ`. VPMULUDQ uses just the low 32 bits of each qword, so the result is identical. The other 7 MACs (offsets 0..24) are unchanged.

Reported in [Consensys/linea-monorepo#3130](https://github.com/Consensys/linea-monorepo/pull/3130), where a workaround was applied caller-side; this PR is the upstream fix.

## Reproducer

mmap a vector that ends exactly at the boundary of a `PROT_NONE` guard page and call `InnerProduct`:

```
unexpected fault address 0x...
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0x... pc=0x...]
github.com/consensys/gnark-crypto/ecc/bls12-377/fr.innerProdVec(...)
	field/asm/element_4w/element_4w_amd64.s:755
```

Reproduced on the unpatched version with a page-aligned 8 MiB slice (`n = 262144`); succeeds with the patch.

## Scope

Audit of every other AVX-512 generator under `internal/generator/field/asm/amd64/` found no similar overread:

- `element_vec_4words_ifma.go` — register-source `VPBROADCASTQ` and full-block (256-byte) `VMOVDQU64` loads aligned to 8-element groups.
- `element_vec_F31*.go` — `VPBROADCASTD` (m32 = 4 bytes), matches F31's 4-byte element size.
- `element_butterfly.go`, `element_mul.go`, `element_reduce.go`, `element_frommont.go` — no memory-broadcast loads.

## Test plan

- [x] Repro program SIGSEGVs on unpatched version, succeeds with patch.
- [x] `go test -short ./ecc/bls12-377/... ./ecc/bls12-381/... ./ecc/bn254/...` passes.
- [x] `go test -short ./ecc/{bn254/fp,bw6-633/fr,grumpkin/fr,secp256k1/fr,stark-curve/fr}` passes.
- [x] `BenchmarkVectorOps/innerProduct` (sizes 16 → 16M, bls12-377): +0.41% geomean (single sample, within noise; one extra 1-µop instruction in a ~41-instruction loop body).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches hand-written/generated AVX-512 assembly in the field arithmetic hot path; while the change is small and intended to be behavior-preserving, mistakes here could cause subtle correctness or performance regressions across multiple curves.
> 
> **Overview**
> Fixes a potential SIGSEGV in the AVX-512 `innerProdVec` loop by avoiding an 8-byte broadcast load at offset `28` (which can read 4 bytes past the 32-byte element). The last multiply-accumulate now uses a new `MAC_LAST` sequence (`VPBROADCASTD` + register `VPMULUDQ`) while keeping the other 7 MACs unchanged.
> 
> Updates the code generator (`element_vec_4words.go`) to emit the new `MAC_LAST` for the final dword and regenerates the per-curve `element_amd64.s` include stubs (hash bump) to pick up the updated shared assembly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68d2af5fa7a68019d738503386337ace5d4ad8f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ivokub (approved):
LGTM! No regression test though, but it is fine

